PR #9, opened 2022-12-30
feat: Fir 17324/move http calls to client

aymeric-dispa:
What it does:

- Fix code check (FIR-18567): 
  - Added step with ` dotnet new globaljson --sdk-version '${{ steps.dotnet-setup.outputs.dotnet-version }}'` to use the version of dotnet installed in the previous step, otherwise the latest version installed will be picked (such as the version shiped with ubuntu)

- Improve/fix integration tests:
  - Added login details so that we can run the tests.
  - Use FIREBOLT_ENGINE_NAME to use created engine
  - Use OneTimeSetUp  & OneTimeTearDown to setup the system engine and remove it only once

- Move HTTP calls from Command to Client(FIR-17324)
  - And removed/refactored some of the code duplicates as the same code to send queries/handle responses was used in 5 different methods
- Encapsulate login and caching login in Client(FIR-17324)
- Fix unit tests (one test was failing)
- Switch using of HttpClient to a Singleton, as suggested by the microsoft docs


FireboltDotNetSdk.Tests/Integration/ExecuteTest.cs
```diff
@@ -57,5 +58,23 @@ public void ExecuteSetEngineTest(string commandText)
...
         }
+
+        [Test]
+        public void ExecuteTestInvalidCredentials()
+        {
+            var connString = $"database={Database};username={Username};password=wrongPassword;endpoint={Endpoint};";
+
+            using var conn = new FireboltConnection(connString);
+
+            try
```
stepansergeevitch, line 69: Can we use `Assert.Throws` and `Assert.ThrowsAsync` in tests, when we want to check that a specific error is thrown? I believe we can fetch exception object to check the message as well.
https://docs.nunit.org/articles/nunit/writing-tests/assertions/classic-assertions/Assert.Throws.html

stepansergeevitch (approved):
Looks good. Left a small comment

PR #12, opened 2023-01-12
feat: fir 18775 add support for new date types

aymeric-dispa:
Ticket: https://packboard.atlassian.net/browse/FIR-18775

### DIFF
--- a/FireboltDotNetSdk.Tests/Integration/ExecuteTest.cs
+++ b/FireboltDotNetSdk.Tests/Integration/ExecuteTest.cs
@@ -120,5 +120,24 @@ namespace FireboltDotNetSdk.Tests
             Assert.That(newMeta.Meta, Is.EqualTo("TimestampTz"));
 
         }
+
+        [Test]
+        public void ExecuteSelectTimestampTzWithUnusualTz()
+        {
+            var connString =
+                $"database={Database};username={Username};password={Password};endpoint={Endpoint};account={Account}";
+
+            using var conn = new FireboltConnection(connString);
+            conn.Open();
+            conn.SetEngine(Engine);
+            var command = conn.CreateCursor();
+            command.Execute("SELECT '2022-05-10 23:01:02.123456 Asia/Calcutta'::timestamptz");
+
+            DateTime dt = DateTime.Parse("2022-05-10 17:31:02.123456Z");
+            NewMeta newMeta = ResponseUtilities.getFirstRow(command.Response);
+            Assert.That(newMeta.Data[0], Is.EqualTo(dt));
+            Assert.That(newMeta.Meta, Is.EqualTo("TimestampTz"));
+
+        }
     }
 }
--- a/FireboltDotNetSdk.Tests/Integration/ExecuteTest.cs
+++ b/FireboltDotNetSdk.Tests/Integration/ExecuteTest.cs
@@ -122,18 +122,19 @@ namespace FireboltDotNetSdk.Tests
         }
 
         [Test]
-        public void ExecuteSelectTimestampTzWithUnusualTz()
+        public void ExecuteSelectTimestampTzWithDifferentFormatTz()
         {
             var connString =
                 $"database={Database};username={Username};password={Password};endpoint={Endpoint};account={Account}";
-
             using var conn = new FireboltConnection(connString);
             conn.Open();
             conn.SetEngine(Engine);
             var command = conn.CreateCursor();
-            command.Execute("SELECT '2022-05-10 23:01:02.123456 Asia/Calcutta'::timestamptz");
+            command.Execute("SET advanced_mode=1");
+            command.Execute("SET time_zone=Asia/Calcutta");
+            command.Execute("SELECT '2022-05-11 23:01:02.123123 Europe/Berlin'::timestamptz");
 
-            DateTime dt = DateTime.Parse("2022-05-10 17:31:02.123456Z");
+            DateTime dt = DateTime.Parse("2022-05-11 21:01:02.123123Z");
             NewMeta newMeta = ResponseUtilities.getFirstRow(command.Response);
             Assert.That(newMeta.Data[0], Is.EqualTo(dt));
             Assert.That(newMeta.Meta, Is.EqualTo("TimestampTz"));
--- a/FireboltDotNetSdk.Tests/Integration/ExecuteTest.cs
+++ b/FireboltDotNetSdk.Tests/Integration/ExecuteTest.cs
@@ -67,7 +67,7 @@ namespace FireboltDotNetSdk.Tests
             FireboltException exception = Assert.Throws<FireboltException>(() => conn.Open());
             Assert.IsTrue(exception.Message.Contains("403") || exception.Message.Contains("429"));
         }
-        
+
         [Test]
         public void ExecuteSelectTimestampNtz()
         {
@@ -80,11 +80,11 @@ namespace FireboltDotNetSdk.Tests
             var command = conn.CreateCursor();
             command.Execute("SELECT '2022-05-10 23:01:02.0'::timestampntz");
             DateTime dt = new DateTime(2022, 5, 10, 23, 1, 2, 0);
-            NewMeta newMeta = getFirstRow(command.Response);
+            NewMeta newMeta = ResponseUtilities.getFirstRow(command.Response);
             Assert.AreEqual("TimestampNtz", newMeta.Meta);
             Assert.AreEqual(dt, newMeta.Data[0]);
         }
-        
+
         [Test]
         public void ExecuteSelectTimestampPgDate()
         {
@@ -96,12 +96,12 @@ namespace FireboltDotNetSdk.Tests
             conn.SetEngine(Engine);
             var command = conn.CreateCursor();
             command.Execute("SELECT '2022-05-10'::pgdate");
-            DateTime dt = new DateTime(2022, 5, 10,0,0, 0);
-            NewMeta newMeta = getFirstRow(command.Response);
+            DateTime dt = new DateTime(2022, 5, 10, 0, 0, 0);
+            NewMeta newMeta = ResponseUtilities.getFirstRow(command.Response);
             Assert.AreEqual("Date", newMeta.Meta);
             Assert.AreEqual(DateOnly.FromDateTime(dt), newMeta.Data[0]);
         }
-        
+
         [Test]
         public void ExecuteSelectTimestampTz()
         {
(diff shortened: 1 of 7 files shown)


FireboltDotNetSdk.Tests/Integration/ExecuteTest.cs
```diff
@@ -68,5 +67,58 @@ public void ExecuteTestInvalidCredentials()
...
+        {
+            var connString =
+                $"database={Database};username={Username};password={Password};endpoint={Endpoint};account={Account}";
+
+            using var conn = new FireboltConnection(connString);
+            conn.Open();
+            conn.SetEngine(Engine);
+            var command = conn.CreateCursor();
+            command.Execute("SELECT '2022-05-10 23:01:02.0'::timestampntz");
+            DateTime dt = new DateTime(2022, 5, 10, 23, 1, 2, 0);
```
ptiurin, line 82: What about the microsecond precision of timestamp(n)tz? We should check that .net sdk can deal with those without losing precision.

FireboltDotNetSdk.Tests/Integration/ExecuteTest.cs
```diff
@@ -68,5 +67,58 @@ public void ExecuteTestInvalidCredentials()
...
+            var connString =
+                $"database={Database};username={Username};password={Password};endpoint={Endpoint};account={Account}";
+
+            using var conn = new FireboltConnection(connString);
+            conn.Open();
+            conn.SetEngine(Engine);
+            var command = conn.CreateCursor();
+            command.Execute("SELECT '2022-05-10 23:01:02.0 Europe/Berlin'::timestamptz");
+
+            DateTime dt = DateTime.Parse("2022-05-10 21:01:02Z");
```
ptiurin, line 117: I thought the time difference between Berlin and UTC is 1 hour? Why is there 2 hours here? Can we set local timezone so this is not platform-dependent?

FireboltDotNetSdk.Tests/Integration/ExecuteTest.cs
aymeric-dispa: milliseconds for timestamptz and timestampntz are already tested in [FireboltCommandTest.cs](https://github.com/firebolt-db/firebolt-net-sdk/pull/12/files#diff-8909c06177d5929e9f73c9361e4064bafeb1d53685c2db88eb64149c97238666) - I didn't want to copy/paste the exact same scenario

FireboltDotNetSdk.Tests/Integration/ExecuteTest.cs
aymeric-dispa: 2022-05-10 21:01:02.0+00 is what is returned by the server. Since I'm creating the time with the timezone (Z), it's actually not platform-dependent (time is not converted to the platform timezone and remains in Zulu time)

aymeric-dispa: I asked postgres and it's expected
image

aymeric-dispa: I just double checked but it's actually correct according to timezoneconverter.com 😅

FireboltDotNetSdk.Tests/Integration/ExecuteTest.cs
ptiurin: I still think it's worth testing the full precision, since the other one is a unit test. It can be a different value but in an integration test it would verify everything works end-to-end.
Also in a linked test precision is lost? https://github.com/firebolt-db/firebolt-net-sdk/pull/12/files#diff-8909c06177d5929e9f73c9361e4064bafeb1d53685c2db88eb64149c97238666R224

aymeric-dispa: It's added at the next lime as the constructor only permits to specify milliseconds

FireboltDotNetSdk.Tests/Integration/ExecuteTest.cs
ptiurin: For an example I used it gave me a 1-hour offset.
image
I understand now, it's daylight saving time that's affecting it. Setting the full date in converter also gives 2 hour offset.

FireboltDotNetSdk.Tests/Integration/ExecuteTest.cs
ptiurin: Ah, missed that part. Fair enough.

aymeric-dispa: > I still think it's worth testing the full precision, since the other one is a unit test. It can be a different value but in an integration test it would verify everything works end-to-end. 
Just added them 😄

aymeric-dispa:
> Looks good. As @stepansergeevitch mentioned in the meeting, it's also worth checking for unusual timezones like `Asia/Calcutta`, as they seem to break parsing (which is the case in GO for me).

I just added one test just to be sure

ptiurin:
> Add test with tz Asia/Calcutta

Sorry, more context. So by default Firebolt converts the the date to a server timezone, which is UTC, before returning it. This makes the test result the same as any other tz. However, if you do `SET advanced_mode=1; SET time_zone='Asia/Calcutta';` effectively saying that you want time to be returned in that tz, it returns you a half-zone tz notation, which may break things.

aymeric-dispa:
> > Add test with tz Asia/Calcutta
> 
> Sorry, more context. So by default Firebolt converts the the date to a server timezone, which is UTC, before returning it. This makes the test result the same as any other tz. However, if you do `SET advanced_mode=1; SET time_zone='Asia/Calcutta';` effectively saying that you want time to be returned in that tz, it returns you a half-zone tz notation, which may break things.

Aaah ok ! Yup looks like there was the same issue (just pushed a fix and added a test)

FireboltDotNetSdk.Tests/Integration/ExecuteTest.cs
```diff
@@ -68,5 +67,78 @@ public void ExecuteTestInvalidCredentials()
...
+                $"database={Database};username={Username};password={Password};endpoint={Endpoint};account={Account}";
+            using var conn = new FireboltConnection(connString);
+            conn.Open();
+            conn.SetEngine(Engine);
+            var command = conn.CreateCursor();
+            command.Execute("SET advanced_mode=1");
+            command.Execute("SET time_zone=Asia/Calcutta");
+            command.Execute("SELECT '2022-05-11 23:01:02.123123 Europe/Berlin'::timestamptz");
+
+            DateTime dt = DateTime.Parse("2022-05-11 21:01:02.123123Z");
```
ptiurin, line 137: Is this correct? `Asia/Calcutta` 's offset is 5:30 from GMT. 
image

ptiurin (commented):
I think it would be worthwhile to add a unit test where server response is `2022-05-10 21:01:02.12345+00:00` so we're covering this format as well.

FireboltDotNetSdk.Tests/Integration/ExecuteTest.cs
aymeric-dispa: When you do SELECT '2022-05-11 23:01:02.123123 Europe/Berlin', you actually get the server response in the Asia/calcutta timezone.
The response from the server is: "2022-05-12 02:31:02.123123+05:30", which in zulu time is as below
image

I also gave it a go with Postgres and I'm getting the same result (it's just showing it for my current tz):
image

FireboltDotNetSdk.Tests/Unit/FireboltCommandTest.cs
```diff
@@ -188,5 +189,64 @@ public void SetParamListDatesTest()
             var expect = commandText.ToString("yyyy-MM-dd HH:mm:ss");
             Assert.That("'" + expect + "'", Is.EqualTo(result));
         }
+
+        [Test]
+        public void TimestampTzTest()
```
stepansergeevitch, line 194: nit: Consider using nunit testcase https://docs.nunit.org/articles/nunit/writing-tests/attributes/testcase.html

FireboltDotNetSdk.Tests/Unit/FireboltCommandTest.cs
```diff
@@ -188,5 +189,64 @@ public void SetParamListDatesTest()
...
+        {
+            var responseWithTimestampTz =
+                "{\"query\":{\"query_id\": \"1739956EA85D7645\"},\"meta\":[{\"name\": \"CAST('2022-05-10 23:01:02.12345 Europe\\/Berlin' AS timestamptz)\",\"type\": \"TimestampTz\"}],\"data\":[[\"2022-05-10 21:01:02.12345+00\"]],\"rows\": 1,\"statistics\":{\"elapsed\": 0.001312549,\"rows_read\": 1,\"bytes_read\": 1,\"time_before_execution\": 0.000553908,\"time_to_execute\": 0.000173431,\"scanned_bytes_cache\": 0,\"scanned_bytes_storage\": 0}}";
```
stepansergeevitch, line 197: I would also suggest adding tests for non-unix timestamp values for timestamptz and timestampntz values.

FireboltDotNetSdk.Tests/Unit/FireboltCommandTest.cs
stepansergeevitch: Also, please add tests for timezone values with seconds `+11:11:11`, since we can get those as well.
For example, with `time_zone=Asia/Calcutta` execute `1111-01-05 17:04:42.123456`. The response will be `1111-01-05 17:04:42.123456+05:53:28`, which is actually correct for some reason

FireboltDotNetSdk.Tests/Unit/FireboltCommandTest.cs
aymeric-dispa: Sure added it to https://github.com/firebolt-db/firebolt-net-sdk/blob/bf0c8797bf9ce0cc1b53765365e2cf987bd6cb14/FireboltDotNetSdk.Tests/Unit/FireboltCommandTest.cs#L241-L260

aymeric-dispa:
> I think it would be worthwhile to add a unit test where server response is `2022-05-10 21:01:02.12345+00:00` so we're covering this format as well.

ok ! Just added the test ExecuteSelectTimestampTzWithTzWithDefaultTz which covers it

FireboltDotNetSdk.Tests/Unit/FireboltCommandTest.cs
aymeric-dispa: I had to change how the ts was handled but it seems ok now - I added the following test
https://github.com/firebolt-db/firebolt-net-sdk/blob/bf0c8797bf9ce0cc1b53765365e2cf987bd6cb14/FireboltDotNetSdk.Tests/Integration/ExecuteTest.cs#L144-L160

stepansergeevitch (approved):
LGTM

ptiurin (approved):
lgtm

PR #13, opened 2023-01-17
fix: Handle value conversion when the type is Nullable(x) and the value to convert is null

aymeric-dispa:
Ticket : https://packboard.atlassian.net/browse/FIR-19100


stepansergeevitch (approved):
Looks good

PR #14, opened 2023-01-17
feat: fir 19080 system engine test improvement

aymeric-dispa:
Ticket: https://packboard.atlassian.net/browse/FIR-19080

### DIFF
--- a/FireboltDotNetSdk.Tests/Integration/SystemEngineTest.cs
+++ b/FireboltDotNetSdk.Tests/Integration/SystemEngineTest.cs
@@ -7,11 +7,9 @@ namespace FireboltDotNetSdk.Tests
     [TestFixture]
     internal class SystemEngineTest : IntegrationTest
     {
-
-
         private FireboltConnection? Connection;
-        private static string EngineName = "system_engine_dotnet_test";
-        private static string DatabaseName = "system_engine_dotnet_test";
+        private static string newEngineName = "system_engine_dotnet_test";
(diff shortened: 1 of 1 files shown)


FireboltDotNetSdk.Tests/Integration/SystemEngineTest.cs
```diff
@@ -35,25 +32,25 @@ public void Cleanup()
...
+                {
+                    cursor.Execute($"DROP ENGINE IF EXISTS {newEngineName}");
+                }
+                catch (System.Exception) { }
                 try
                 {
-                    cursor.Execute($"DROP DATABASE IF EXISTS {DatabaseName}");
+                    cursor.Execute($"DROP DATABASE IF EXISTS {newDatabaseName}");
                 }
                 catch (FireboltException) { };
```
ptiurin, line 48: Should we fail if cleanup is not possible?

ptiurin (approved):
Just one question, otherwise LGTM.

FireboltDotNetSdk.Tests/Integration/SystemEngineTest.cs
aymeric-dispa: I don't want to do assertions in the cleanup method as any failure would leave the method (could play with booleans but I want to keep it short and clean). Instead I will copy that line in a test and I'll keep the cleanup method as is

stepansergeevitch (approved):
lgtm

PR #17, opened 2023-01-19
feat(FIR-15835): Add support for service account

aymeric-dispa:
Ticket: https://packboard.atlassian.net/browse/FIR-15835?search_id=62d4119d-b621-458e-9afb-c6e5c1c0f74a

### DIFF
--- a/FireboltNETSDK/Client/FireRequest.cs
+++ b/FireboltNETSDK/Client/FireRequest.cs
@@ -15,6 +15,8 @@
  */
 #endregion
 
+using Newtonsoft.Json;
+
 namespace FireboltDotNetSdk.Client
 {
     public class FireRequest
@@ -30,12 +32,14 @@ namespace FireboltDotNetSdk.Client
             /// <summary>
             /// Password.
             /// </summary>
-            public string Password { get; set; }
+            [JsonProperty]
+            private string Password { get;}
 
             /// <summary>
             /// Username.
             /// </summary>
-            public string Username { get; set; }
+            [JsonProperty]
+            private string Username { get;}
 
         }
 
--- a/FireboltNETSDK/Client/FireRequest.cs
+++ b/FireboltNETSDK/Client/FireRequest.cs
@@ -19,8 +19,14 @@ namespace FireboltDotNetSdk.Client
 {
     public class FireRequest
     {
-        public class LoginRequest
+        public class UsernamePasswordLoginRequest
         {
+            public UsernamePasswordLoginRequest(string username, string password)
+            {
+                Password = password;
+                Username = username;
+            }
+
             /// <summary>
             /// Password.
             /// </summary>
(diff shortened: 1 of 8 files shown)


FireboltNETSDK/Client/FireRequest.cs
```diff
@@ -32,5 +38,45 @@ public class LoginRequest
             public string Username { get; set; }
```
ptiurin, line 38: Outside of scope, but should we make Username Password private, similar to service account attributes?

ptiurin (approved):
LGTM

FireboltNETSDK/Client/FireRequest.cs
aymeric-dispa: Nope we can't because UsernamePasswordLoginRequest needs to expose the Username and Password (getters are used outside of that class). ClientId and ClientSecret are just used inside the ServiceAccountLoginRequest class so they can be private

ptiurin: I don't see where those are used outside of a class. Also, that doesn't really explain why we can't, but rather how public/private works.

aymeric-dispa: They are during the serialization 😉 . Making those private would 'work' - as the code would still compile, but login would stop working. Serialization is not done within the class itself because it depends on the serialization settings that are part of the Client class.

ptiurin: How does username/password serialisation differ from service account then? We're sending the same information via http in both auth methods, the difference is only the api endpoint.

aymeric-dispa: The content type is actually different (application/x-www-form-urlencoded). We are sending a FormUrlEncodedContent contaning an iterable of keys/values for SA. For login with username/password, the UsernamePasswordLoginRequest is serialized to a json payload (using the JsonSerializer - which uses the getters).

ptiurin: Right, so if we move the serialiser within the class, similarly to what you've done with `GetFormUrlEncodedContent`, we would not have to expose the credentials either, doesn't matter the payload.

ptiurin: I'm not asking to do this within this change, but it's not like we can't or shouldn't do it.

aymeric-dispa: Moving the serializer wouldn't do (we use the method JsonConvert.SerializeObject - which uses the getters - and calling it inside the class wouldn't directly expose private members of that class).
However it looks like `@JsonProperty `does the trick though :) I will update it

ptiurin: Thanks. Yes, that's what I meant. We're not chatgpt - I'm not suggesting just replacing public to private and hoping it works. My suggestion was we can refactor this to make it work and that maybe this would be cleaner in terms of isolation of attributes as well as having two similar classes work somewhat similarly.
Most of the time there's no "can't" and more like infeasible or there's a specific bug/problem it would cause instead. I'd rather have that as an explanation.

PR #23, opened 2023-02-28
refactor: nullability fixes

ptiurin:
Eliminating nullability compiler warnings by clarifying it throughout the code.

Main assumptions here:
* Better throw exception with descriptive type/message when unexpected null is encountered than forcing or assuming it won't happen.
* In tests we first check if nullable attribute is not null until checksing sub-attributes. While notnull assert is not blocking (other asserts will also evaluate) it will be the first one to error out, therefore indicating the actual error a bit more accurately.
* rowCount will return -1 if data does not contain it. This is consistent with the Python SDK.
* AesManaged is deprecated in favour of equivalent Aes.Create()

Tests are passing.

### DIFF
--- a/FireboltNETSDK/Client/FireboltConnection.cs
+++ b/FireboltNETSDK/Client/FireboltConnection.cs
@@ -80,6 +80,7 @@ namespace FireboltDotNetSdk.Client
 
         public override string DataSource => throw new NotImplementedException();
 
+        [System.Diagnostics.CodeAnalysis.AllowNull]
         public override string ConnectionString { get => throw new NotImplementedException(); set => throw new NotImplementedException(); }
 
         /// <summary>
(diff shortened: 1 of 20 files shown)


FireboltNETSDK/Client/FireboltConnection.cs
```diff
@@ -189,14 +191,22 @@ public override async Task<bool> OpenAsync(CancellationToken cancellationToken)
             }
         }
 
-        public GetEngineUrlByEngineNameResponse SetEngine(string? engineName)
+        public GetEngineUrlByEngineNameResponse SetEngine(string engineName)
         {
+            if (Client is null)
```
stepansergeevitch, line 196: Can we make this consistent with line 172? Either with or without `{}`

FireboltNETSDK/Client/FireboltParameter.cs
```diff
@@ -56,7 +56,8 @@ public override DbType DbType
         {
             get
             {
-                var chType = FireboltDbType;
+                // not sure if this is right
+                var chType = FireboltDbType ?? global::FireboltDbType.String;
```
stepansergeevitch, line 60: We have the same behaviour in Python SDK. This shouldn't happen anyways, should it?

stepansergeevitch (approved):
LGTM

FireboltNETSDK/Client/FireboltConnection.cs
ptiurin: done

FireboltNETSDK/Client/FireboltParameter.cs
ptiurin: You're right. What was worrying me here is that this class is used in parametrised queries, but there are checks for the correct type. Also, this represents db type for a given value, if value is not recognised it makes sense for it to default to string.

ptiurin:
Fixing coverage is out of scope of this PR so I'm force merging.

PR #26, opened 2023-03-13
feat(FIR-16885): Move engine setting to connection string

aymeric-dispa:
The goal of this PR is to:

- Enable the passing of engine name as part of the connection string
- Deprecate the method setEngine() and setDefaultEngine()
- Improve exception handling when an engine is not found (fail fast vs setting the url to null and have the connector failing when sending a query to the server).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,4 +29,4 @@ jobs:
     - name: Build project
       id: build
       run: |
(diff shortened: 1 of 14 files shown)


ptiurin (approved):
LGTM

stepansergeevitch (approved):
LGTM

PR #29, opened 2023-04-04
feat: New identity

ptiurin:
Added support for the new identity, extended tests to cover new functionality

### DIFF
--- a/FireboltDotNetSdk.Tests/Integration/ConnectionTest.cs
+++ b/FireboltDotNetSdk.Tests/Integration/ConnectionTest.cs
@@ -14,7 +14,7 @@ namespace FireboltDotNetSdk.Tests
             FireboltConnection Connection = new FireboltConnection(connString);
             Connection.Open();
             var cursor = Connection.CreateCursor();
-            cursor.Execute("SELECT 1");
+            cursor.Execute("SELECT TOP 1 * FROM information_schema.tables");
             Assert.NotNull(cursor.Response);
             NewMeta newMeta = ResponseUtilities.getFirstRow(cursor.Response!);
             Assert.That(newMeta.Data[0], Is.EqualTo(1));
@@ -26,7 +26,7 @@ namespace FireboltDotNetSdk.Tests
             FireboltConnection Connection = new FireboltConnection(connString);
             Connection.Open();
             var cursor = Connection.CreateCursor();
-            cursor.Execute("SELECT 1");
+            cursor.Execute("SELECT TOP 1 * FROM information_schema.tables");
             Assert.NotNull(cursor.Response);
             NewMeta newMeta = ResponseUtilities.getFirstRow(cursor.Response!);
             Assert.That(newMeta.Data[0], Is.EqualTo(1));
@@ -38,7 +38,7 @@ namespace FireboltDotNetSdk.Tests
             FireboltConnection Connection = new FireboltConnection(connString);
             Connection.Open();
             var cursor = Connection.CreateCursor();
-            cursor.Execute("SELECT 1");
+            cursor.Execute("SELECT TOP 1 * FROM information_schema.tables");
             Assert.NotNull(cursor.Response);
             NewMeta newMeta = ResponseUtilities.getFirstRow(cursor.Response!);
             Assert.That(newMeta.Data[0], Is.EqualTo(1));
--- a/FireboltDotNetSdk.Tests/Integration/ConnectionTest.cs
+++ b/FireboltDotNetSdk.Tests/Integration/ConnectionTest.cs
@@ -40,5 +40,17 @@ namespace FireboltDotNetSdk.Tests
                     () => cursor.Execute("SELECT TOP 1 * FROM information_schema.tables"));
             }
         }
+
+        [Test]
+        public void InvalidAccountConnectTest()
+        {
+            var connString = $"clientid={ClientId};clientsecret={ClientSecret};account=non-existing-account-123;env={Env}";
+            FireboltConnection Connection = new FireboltConnection(connString);
+
+            FireboltException? exception = Assert.Throws<FireboltException>(
+                        () => Connection.Open());
+            Assert.NotNull(exception);
+            Assert.That(exception!.Message, Does.Contain("Account with name non-existing-account-123 was not found"));
+        }
     }
 }
--- a/FireboltDotNetSdk.Tests/Integration/ConnectionTest.cs
+++ b/FireboltDotNetSdk.Tests/Integration/ConnectionTest.cs
@@ -32,7 +32,7 @@ namespace FireboltDotNetSdk.Tests
             Assert.That(newMeta.Data[0], Is.EqualTo(1));
         }
         [Test]
-        public void ConnectuserEngineNoDbTest()
+        public void ConnectUserEngineNoDbTest()
         {
             var connString = $"clientid={ClientId};clientsecret={ClientSecret};account={Account};env={Env};engine={EngineName}";
             FireboltConnection Connection = new FireboltConnection(connString);
--- a/FireboltDotNetSdk.Tests/Integration/ConnectionTest.cs
+++ b/FireboltDotNetSdk.Tests/Integration/ConnectionTest.cs
@@ -1,4 +1,5 @@
 using FireboltDotNetSdk.Client;
+using FireboltDotNetSdk.Exception;
 
 
 namespace FireboltDotNetSdk.Tests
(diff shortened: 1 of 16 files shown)


README.md
```diff
@@ -63,7 +63,7 @@ Execute command
 
 ```cs
 
-            var connString = $"database={_database};username={_username};password={_password};endpoint={_endpoint};";
+            var connString = $"database={_database};clientid={_clientId};clientsecret={_clientSecret};endpoint={_endpoint};";
```
ptiurin, line 66: I know it was in the old version of the readme as well, but where does `_variable` come from? I think to eliminate any potential confusion we should just change all those variables to the same. e.g. `clientId`, `clientSecret`, etc.

FireboltDotNetSdk.Tests/Integration/ConnectionTest.cs
```diff
@@ -0,0 +1,47 @@
...
+            Assert.That(newMeta.Data[0], Is.EqualTo(1));
+        }
+        [Test]
+        public void ConnectNoEngineWithDatabaseTest()
+        {
+            var connString = $"database={Database};clientid={ClientId};clientsecret={ClientSecret};account={Account};env={Env}";
+            FireboltConnection Connection = new FireboltConnection(connString);
+            Connection.Open();
+            var cursor = Connection.CreateCursor();
+            cursor.Execute("SELECT 1");
```
ptiurin, line 29: Same as in some other connector, I think it's worth testing relevant sql for each test case. e.g. `SELECT TOP 1 * FROM information_schema.tables` would only work if we have a database context. etc.

FireboltDotNetSdk.Tests/Integration/ExecuteTest.cs
ptiurin, line 1: Probably also worth adding a test when account is not provided/invalid account.

FireboltNETSDK/FireboltClient.cs
```diff
@@ -152,19 +132,41 @@ public Task<GetEngineUrlByEngineNameResponse> GetEngineUrlByEngineId(string? eng
...
-        var setParam = setParamList.Aggregate(string.Empty, (current, item) => current + "&" + item);
-        var urlBuilder = new StringBuilder();
-        urlBuilder.Append("https://").Append(engineEndpoint).Append("?database=").Append(databaseName)
-            .Append(setParam).Append("&output_format=JSON_Compact");
-        return await SendAsync<string>(HttpMethod.Post, urlBuilder.ToString(), query, "text/plain", true, cancellationToken);
+        var setParams = setParamList.Aggregate(string.Empty, (current, item) => current + "&" + item);
+        var urlBuilder = new UriBuilder(engineEndpoint)
+        {
+            Scheme = "https",
+            Port = -1
```
ptiurin, line 146: Do we need to provide a port here?

ptiurin (commented):
Some suggestions, otherwise LGTM. Can't approve my own PR. :)

FireboltNETSDK/Client/FireboltConnection.cs
```diff
@@ -158,74 +181,92 @@ public override DataTable GetSchema(string collectionName, string?[] restriction
...
+                _isSystem = false;
+            }
             OnSessionEstablished();
             return EngineUrl != null;
         }
 
+        private string? GetEngineDatabase(string engineName)
+        {
+            var query = "SELECT attached_to FROM information_schema.engines " +
+                        $"WHERE engine_name='{engineName}'";
```
alexradzin, line 214: can we use prepared statement here?

FireboltNETSDK/Client/FireboltConnection.cs
```diff
@@ -158,74 +181,92 @@ public override DataTable GetSchema(string collectionName, string?[] restriction
...
-        private string? GetEngineUrlByEngineName(string engineName, FireboltClient client)
+        private string? GetEngineUrl(string engineName, string database)
         {
-            try
+            var query = "SELECT engs.url, engs.attached_to, dbs.database_name, status " +
+                        "FROM information_schema.engines as engs " +
+                        "LEFT JOIN information_schema.databases as dbs " +
+                        "ON engs.attached_to = dbs.database_name " +
+                        $"WHERE engs.engine_name = '{engineName}'";
+            var cursor = CreateCursor();
```
alexradzin, line 247: can we use prepared statement here? (same as previous)

FireboltNETSDK/Client/FireboltConnectionSettings.cs
```diff
@@ -40,29 +42,71 @@ public sealed class FireboltConnectionSettings
...
+            "Configuration error: environment " +
+            $"{env} and endpoint {endpoint} are incompatible"
+        );
+            }
+            if (env == null && endpoint_env != null)
+            {
+                env = endpoint_env;
+            }
+            System.Console.WriteLine($"Endpoint: {endpoint}");
+            System.Console.WriteLine($"Env: {env}");
```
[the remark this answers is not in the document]
stepansergeevitch: Yup, forgot to remove

FireboltNETSDK/Utils/Constant.cs
```diff
@@ -52,5 +52,7 @@ static class Constant
 
         public static string PROVIDERS_URL = "/compute/v1/providers";
         public static string REGIONS_URL = "/compute/v1/regions";
+
+        public static string ACCOUNT_SYSTEM_ENGINE_URL = "/web/v3/account/{0}/engineUrl";
```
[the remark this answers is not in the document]
stepansergeevitch: C# string format has no easy way to do this except for using third-party libraries. Yet, since we have only one parameter mostly in these strings, we can use `.Replace`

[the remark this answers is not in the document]
stepansergeevitch: Looking further into it, Replace doesn't seem to be the best fit here. I think we should leave ordered parameters here until a better solution is introduced in C#

README.md
stepansergeevitch: This makes total sense, didn't notice it. Also, there are some more updated to make in Readme, like removing `SetEngine` example since we don't have that function anymore

FireboltNETSDK/FireboltClient.cs
stepansergeevitch: We set port to -1, otherwise it will automatically add it to the resulting string, like `eng.uuid.us-east-1.dev.firebolt.io:443`, and that doesn't work in some cases.

FireboltDotNetSdk.Tests/Integration/ConnectionTest.cs
```diff
@@ -0,0 +1,47 @@
...
+            Assert.NotNull(cursor.Response);
+            NewMeta newMeta = ResponseUtilities.getFirstRow(cursor.Response!);
+            Assert.That(newMeta.Data[0], Is.EqualTo(1));
+        }
+    }
+}
```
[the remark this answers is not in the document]
stepansergeevitch: Sure, that makes sense!

FireboltDotNetSdk.Tests/Integration/SystemEngineTest.cs
```diff
@@ -171,6 +171,7 @@ private void VerifyEngineStatus(FireboltCommand cursor, string engineName, strin
         }
 
         [Test]
+        [Ignore("Start stop command are flaky now, enable this test when they become stable")]
```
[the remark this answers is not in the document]
stepansergeevitch: I think we can remove the Ignore now

FireboltNETSDK/Client/FireboltConnection.cs
```diff
@@ -158,74 +181,95 @@ public override DataTable GetSchema(string collectionName, string?[] restriction
...
+                _database = database;
+                _isSystem = false;
+            }
             OnSessionEstablished();
             return EngineUrl != null;
         }
 
+        private string? GetEngineDatabase(string engineName)
+        {
+            var query = "SELECT attached_to FROM information_schema.engines " +
```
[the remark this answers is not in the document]
stepansergeevitch: I agree

FireboltNETSDK/Client/FireboltConnectionSettings.cs
```diff
@@ -40,29 +42,69 @@ public sealed class FireboltConnectionSettings
...
+        }
+
+        static string? ExtractEndpointEnv(string endpoint)
+        {
+            // Remove http:// or https:// prefix if present
+            endpoint = endpoint.Split("://")[^1];
+            var endpoint_parts = endpoint.Split(".");
+            // Verify the expected 
+            if (
+            endpoint_parts.Length == 4 &&
```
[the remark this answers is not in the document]
stepansergeevitch: Sure

alexradzin (approved):

PR #36, opened 2023-08-14
fix: Fix sdk version to conform to a standard

stepansergeevitch:
Since a new `1.0.0-alpha` version was created, Github actions bot has automatically updated the version package to `1.0.0-alpha`. But this version value is considered invalid by dotnet, which causes the build to fail. Setting the version to `1.0.0` to fix the build


ptiurin (changes requested):
Hold on this one. Let's discuss in daily. .net is supposed to support -alpha release

alexradzin (approved):

ptiurin (approved):
Lgtm, we'd need to publish to Nuget manually as the workflow that does it also bumps the versions.

PR #37, opened 2023-08-27
feat: More metadata

alexradzin:


FireboltNETSDK/Client/FireboltConnection.cs
```diff
@@ -98,12 +99,53 @@ public string ClientId
         /// <returns>The state of the connection.</returns>
         public override ConnectionState State => _connectionState.State;
 
-        public override string ServerVersion => throw new NotImplementedException();
+        public override string ServerVersion
+        {
+            get => (string?)getOneLine("SELECT VERSION()")?[0] ?? string.Empty;
```
stepansergeevitch, line 104: Does it make sense to cache the value here?

FireboltNETSDK/Client/FireboltConnection.cs
```diff
@@ -316,6 +384,7 @@ public override void Close()
         {
             _connectionState.State = ConnectionState.Closed;
             Client = null;
+            _isSystem = true;
```
stepansergeevitch, line 387: Why do we set connection to system when we close it?

stepansergeevitch (approved):
A couple of comments, LGTM otherwise

FireboltNETSDK/Client/FireboltConnection.cs
alexradzin: We have to restore the state. Otherwise we cannot open connection again. Take a look on line `if (_accountId == null && Account != null && _isSystem)`.
I spent a lot of time debugging this.

PR #38, opened 2023-08-30
feat: (FIR-16888) fetch token when 401 for new identity

alexradzin:


FireboltNETSDK/FireboltClient.cs
```diff
@@ -170,7 +176,7 @@ public async Task<GetSystemEngineUrlResponse> GetSystemEngineUrl(string accountN
         urlBuilder.Query = queryStr.ToString();
         var url = urlBuilder.Uri.ToString();
 
-        return await SendAsync<string>(HttpMethod.Post, url, query, "text/plain", true, cancellationToken);
+        return await SendAsync<string>(HttpMethod.Post, url, query, _textContentType, true, cancellationToken, true);
```
ptiurin, line 179: It's not clear what `true` argument represents here (both true's in fact)? Can we either create a variable or use named parameters (if c# supports it) so those are a bit more obvious when going through the code?

FireboltNETSDK/FireboltClient.cs
```diff
@@ -293,6 +298,12 @@ private async Task<T> SendAsync<T>(HttpMethod method, string uri, HttpContent? c
             }
             else
             {
+                if (needsAccessToken && response.StatusCode == HttpStatusCode.Unauthorized && retryUnauthorized)
+                {
+                    //If we need an access token and the token is invalid (401), we try to re-establish the connection once.
+                    await EstablishConnection(true);
```
ptiurin, line 304: Same as above, `await EstablishConnection(forceTokenRefresh=true);` while redundant is more readable.

FireboltNETSDK/FireboltClient.cs
```diff
@@ -311,21 +322,65 @@ private async Task<T> SendAsync<T>(HttpMethod method, string uri, HttpContent? c
...
     {
         if (string.IsNullOrEmpty(_token))
         {
-            throw new FireboltException("The Access token is null or empty - EstablishConnection must be called");
+            _token = await EstablishConnection();
         }
         request.Headers.Add("Authorization", "Bearer " + _token);
     }
 
+    private static string ConvertToString(object? value, CultureInfo cultureInfo)
```
ptiurin, line 343: Is this method used anywhere? I can't find references to it.

FireboltNETSDK/FireboltClient.cs
alexradzin: done

FireboltNETSDK/FireboltClient.cs
alexradzin: You are right, this is not in use. This function was added in April

```
alexradzin@LenovoT14Firebolt:~/proj/firebolt-db/firebolt-net-sdk$ git blame ./FireboltNETSDK/FireboltClient.cs | grep ConvertToString
a4d7da2b (Aymeric           2023-04-03 17:15:27 +0100 343)     private static string ConvertToString(object? value, CultureInfo cultureInfo)
a4d7da2b (Aymeric           2023-04-03 17:15:27 +0100 373)                         return string.Join(",", array.Select(o => ConvertToString(o, cultureInfo)));

```

FireboltNETSDK/FireboltClient.cs
```diff
@@ -244,36 +248,37 @@ public async Task<GetSystemEngineUrlResponse> GetSystemEngineUrl(string accountN
     private async Task<T> SendAsync<T>(HttpMethod method, string uri, string? body, bool requiresAuth,
         CancellationToken cancellationToken)
     {
-        return await SendAsync<T>(method, uri, body, "application/json", requiresAuth, cancellationToken);
```
ptiurin, line 247: Same as before with the unnamed true parameter. Please look through the PR to identify other cases.

ptiurin (approved):

PR #41, opened 2023-09-10
ci: fix test coverage presentation

alexradzin:


alexradzin:
> Can we get some info on why is this changed? Failing below 80% coverage on new code is in line which the CI setup we have in other repos.

Here is the motivation: get real test coverage on each commit/PR. 
What prevented it to happen? The parameter `only_changed_files=true`. What does it mean? It means that the test coverage is measured only on changed files. If for example overall coverage is 80% but we change 2 lines in relatively big file that is is not covered at all. Even if we add good tests that validate those 2 lines (so they are 100% covered) the coverage of this file is almost 0%. This is what we see in build and therefore the build "fails" due to low coverage. IMHO `only_changed_files: false` reflects the overall coverage better.    

`fail_below_threshold: true` prevents build from running next actions if one action failed. In our case if failing of stage `Pull request / unit-tests / coverage (pull_request)` (due to low code coverage) prevents the next stage `Pull request / coverage (pull_request) — coverage` to run and it is marked as failed. IMHO there are idempotent actions and therefore failure of action N should not prevent action N+1 to run. If action N+1 fails itself it will be marked as failed.  

These are the reasons to set the values as following:
```
 fail_below_threshold: false
 only_changed_files: false
```

ptiurin:
>  If for example overall coverage is 80% but we change 2 lines in relatively big file that is is not covered at all. Even if we add good tests that validate those 2 lines (so they are 100% covered) the coverage of this file is almost 0%. 

But the opposite would also be incorrect, if we have 80% overall coverage and the file is 100% covered and two more lines are added with no coverage it will pass. 
 
> `fail_below_threshold: true` prevents build from running next actions if one action failed. In our case if failing of stage `Pull request / unit-tests / coverage (pull_request)` (due to low code coverage) prevents the next stage `Pull request / coverage (pull_request) — coverage` to run and it is marked as failed.

I believe those are the result of the same action, so if coverage passes then "top-level coverage" will be marked as successful. The rest of the actions are running independently of this:
image


IMO, disabling test coverage is not the solution here. SonarCloud has support for dotnet applications, it can report coverage on changed **lines** of code, meaning if 10 lines were modified and 8 were covered by a test it will pass: https://docs.sonarcloud.io/advanced-setup/ci-based-analysis/sonarscanner-for-net/  We should look into using this instead.

alexradzin:
I do not want to disable test coverage. I want to make it usable. What happens right now? Each build  fails because there are a lot of files that do not have enough test coverage. I want to see progress, i.e. I want to see that test coverage is growing up even if some files are not covered right now. One day (probably in a month or two) when the coverage of each single file will exceed 80% we will try  to restore the value of these properties to follow the "fail fast" principal. Otherwise I even do not have a tool to find what files lack tests: they even do not appear in the list displayed by GitHub.

ptiurin:
> I want to see progress, i.e. I want to see that test coverage is growing up even if some files are not covered right now.
> Otherwise I even do not have a tool to find what files lack tests: they even do not appear in the list displayed by GitHub.

For both of those points, see my previous comment about the SonarCloud integration.

Example: https://sonarcloud.io/project/overview?id=firebolt-db_firebolt-python-sdk (you may need to request access in #firebolt-it)
Some features that you're looking for:
image
image

As of now this change has no real benefit apart from spamming each PR with a full list of files that are not covered with tests. If you want to know the full coverage you can run `dotnet test --filter "Category!=Integration" /p:AltCover=true` locally. Even better, SonarCloud integration will make this whole thing obsolete so we won't have to have the coverage action.

PR #42, opened 2023-09-21
refactor: Refactoring

alexradzin:


FireboltNETSDK/Client/FireboltCommand.cs
```diff
@@ -187,41 +218,15 @@ public string GetParamQuery(string commandText)
...
-                                var intValue = (int)item.Value;
-                                verifyParameters = intValue.ToString();
-                                break;
-                            case "Int64":
-                                var longValue = (long)item.Value;
-                                verifyParameters = longValue.ToString();
-                                break;
-                        }
+                        verifyParameters = ((IConvertible)item.Value).ToString(new CultureInfo("en-US", false));
                     }
```
ptiurin, line 230: Can you explain what does this do and why is it being removed please?

FireboltNETSDK/Client/FireboltDataReader.cs
```diff
@@ -0,0 +1,473 @@
...
+                case bool b: return (bool)b;
+                case float f: return (float)f != 0;
+                case double d: return (double)d != 0;
+                case decimal d: return (decimal)d != 0;
+                case byte b: return (byte)b != 0;
+                case short s: return (short)s != 0;
+                case int i: return (int)i != 0;
+                case long l: return (long)l != 0;
+                case string s: return bool.Parse(s);
+                default: throw new InvalidCastException($"Cannot cast ({value.GetType()}){value} to boolean");
```
ptiurin, line 116: As mentioned on CR, let's verify that this is the expected behaviour and we shouldn't just throw an error if the value is not bool.

FireboltNETSDK/Client/FireboltDataReader.cs
```diff
@@ -0,0 +1,473 @@
...
+        }
+
+
+        // public virtual Task<ReadOnlyCollection<DbColumn>> GetColumnSchemaAsync(CancellationToken cancellationToken = default);
+
+        //[EditorBrowsable(EditorBrowsableState.Never)]
+        // public override DbDataReader GetData(int ordinal)
+        // {
+        //     return this;// TODO
+        // }
```
ptiurin, line 182: comment

FireboltNETSDK/Client/FireboltDataReader.cs
```diff
@@ -0,0 +1,473 @@
...
+        }
+
+        public override int GetOrdinal(string name)
+        {
+            return _queryResult?.Meta.FindIndex(m => m.Name == name) ?? throw new IndexOutOfRangeException($"Cannot find index for column {name}");
+        }
+
+        public override DataTable? GetSchemaTable()
+        {
+            return _fullTableName == null ? null : new DataTable(_fullTableName); // TODO split full table name
```
ptiurin, line 350: Can you elaborate on this TODO? Nice to have or something that was missed?

FireboltNETSDK/Client/FireboltDataReader.cs
```diff
@@ -0,0 +1,473 @@
...
+
+        public override bool NextResult()
+        {
+            throw new FireboltException("Batch operations are not supported");
+        }
+
+        public override Task<bool> NextResultAsync(CancellationToken cancellationToken)
+        {
+            return NextResultAsync();
+        }
```
ptiurin, line 441: Should those both work the same way, throwing an exception?

FireboltNETSDK/Client/FireboltParameter.cs
```diff
@@ -12,39 +13,53 @@ namespace FireboltDotNetSdk.Client
...
+            { typeof(string), DbType.String },
+            { typeof(TimeOnly), DbType.Time },
+            { typeof(ushort), DbType.UInt16 },
+            { typeof(uint), DbType.UInt32 },
+            { typeof(ulong), DbType.UInt64 },
+            { typeof(DateTimeOffset), DbType.DateTimeOffset },
+        };
+
+        // This constant is copied from Microsoft's System.Data.SqlTypes.SQLDecimal
+        private const uint DBL_DIG = 17; // Max decimal digits of double
```
ptiurin, line 47: Should this be Firebolt's value instead? We should be clear the precision is 15 and extra digits will be truncated.

FireboltNETSDK/Client/FireboltParameter.cs
```diff
@@ -276,32 +222,19 @@ internal static string TrimParameterName(string parameterName)
...
-            public int? ArrayRank => _parameter._forcedArrayRank;
-
-            public ParameterColumnTypeDescriptorAdapter(FireboltParameter parameter)
+            if (value == null)
             {
-                _parameter = parameter;
+                return 0;
             }
+            //TODO: improve this: make size discovery dependent on value type
+            return (value.ToString() ?? string.Empty).Length;
```
ptiurin, line 232: I feel like this could be potentially misleading. Should we just throw an error for the time being if getsize is called on a non-null value?

FireboltNETSDK/Utils/Types.cs
```diff
@@ -331,30 +354,44 @@ public static IEnumerable<NewMeta> ParseJsonResponse(string? response)
...
             catch (System.Exception e)
             {
                 throw new FireboltException("Error while parsing response", e);
             }
         }
+
+
+        public static IEnumerable<NewMeta> ProcessQueryResult(QueryResult? result)
+        {
+            if (result == null) throw new FireboltException("Unable to parse data to JSON");
```
ptiurin, line 368: IMO, the error about parsing should be next to the parsing function, not the one that makes sense of data afterwards. This method does not know about parsing or JSON. Maybe later on we'll have tab-delimited data instead. 
I'd move this exception to ConvertToDate and if a null passed in here would change the message to something like "Can't process null values".

FireboltNETSDK/Client/FireboltParameterCollection.cs
```diff
@@ -29,158 +34,80 @@ public sealed class FireboltParameterCollection : DbParameterCollection, IList<F
...
-        /// <param name="parameterName">The name of the parameter.</param>
-        /// <param name="value">The value of the paramter.</param>
-        /// <returns>A new <see cref="FireboltParameter"/> added to the collection.</returns>
-        public FireboltParameter AddWithValue(string parameterName, object? value)
-        {
-            var parameter = new FireboltParameter(parameterName) { Value = value };
+            DbParameter parameter = typeof(DbParameter).IsAssignableFrom(value.GetType()) ? (DbParameter)value : new FireboltParameter(FireboltParameter.defaultParameterName, value);
             Add(parameter);
-            return parameter;
+            return _parameters.Count - 1;
```
ptiurin, line 39: Why do we do `-1` here?

FireboltDotNetSdk.Tests/Integration/ConnectionTest.cs
```diff
@@ -180,29 +188,40 @@ public void SetDatabaseUsingConnectionStringFirstGoodThenWrong()
...
-            Connection.Open();
-            assertSelectOne(Connection.CreateCursor());
-            Assert.Throws<E>(() => Connection.ConnectionString = connString2);
+            DbConnection connection = new FireboltConnection(connString1);
+            connection.Open();
+            assertSelectOne(connection.CreateCommand());
+            Assert.Throws<E>(() => connection.ConnectionString = connString2);
+        }
+
+        private void failingSelect1(DbConnection connection)
```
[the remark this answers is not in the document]
alexradzin: `assertSelectOne()` is a function that implements successful scenario. `failingSelect1` implements failing scenario, so I do not think there is a reason to unify the names. However you are right that both suffixes `One` and `1` are redundant, so I removed them.  I this is what you meant, sorry for miss-understanding.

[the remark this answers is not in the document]
ptiurin: Yeah, I meant the names of those functions. Just easier to search for one or another.

FireboltNETSDK/Client/FireboltCommand.cs
alexradzin: You are right, this code is not obvious, so I added comment to code. If you want to see the reason see the full page of code that was in previous version and is replaced by this single line.

FireboltNETSDK/Client/FireboltParameter.cs
alexradzin: Fixed and added comment.

FireboltNETSDK/Utils/Types.cs
alexradzin: I moved this validation to other implementation of this method that accepts JSON as string. Moreover method `ProcessQueryResult(QueryResult? result)` is now private and I have removed the only one reference to it that in turn was not in use.

FireboltNETSDK/Client/FireboltParameterCollection.cs
alexradzin: Because according to the documentation of this method declared in `DbParameterCollection` it returns "The index of the System.Data.Common.DbParameter object in the collection."
The index is zero based, that means that we should subtract 1 from the collection size.

FireboltNETSDK/Client/FireboltParameter.cs
alexradzin: I think implementation of this function is reasonable. If value is null it returns 0. Otherwise it returns size of its string representation. But if the string representation is null itself it is replaced by empty string and therefore 0 will be returned. 

So, this function can be re-written as following:

```
if (value == null)
{
    return 0;
}
string? str value.ToString();
return str == null ? 0 : str.Length;
```

I also added comment 
```
//TODO: improve this: make size discovery dependent on value type
```
that means that I am planning to improve this code.

ptiurin: I understand what it does and what TODO is. I don't agree with the current implementation. Size is supposed to represent a number of bytes for non-string types ([doc](https://learn.microsoft.com/en-us/dotnet/api/system.data.common.dbparameter.size?view=net-7.0#remarks)) so returning a length of their string representation is misleading. We can't just change logic on a whim, even if we add comments.

FireboltNETSDK/Client/FireboltDataReader.cs
alexradzin: fixed

FireboltNETSDK/Client/FireboltParameter.cs
alexradzin: implemented

FireboltNETSDK/Client/FireboltDataReader.cs
alexradzin: done

FireboltNETSDK/Client/FireboltDataReader.cs
alexradzin: Yes, this is up to implementation. Please see the snapshot from the MySQL driver:

```
			return (boolVal == -1 || boolVal > 0);
		case Types.BIT:
		case Types.TINYINT:
		case Types.SMALLINT:
		case Types.INTEGER:
		case Types.BIGINT:
		case Types.DECIMAL:
		case Types.NUMERIC:
		case Types.REAL:
		case Types.FLOAT:
		case Types.DOUBLE:
			boolVal = getLong(columnIndex, false);

```

ptiurin: I understood how, but I had questions to why. To me it sounds counterintuitive to request an incorrect type from the specific type returned from the database and get a result.
I'm not sure why you're using an example in different language and interface when there are opensource c# implementations for many databases. Here's postgres, which we're basing our SQL syntax on - https://github.com/npgsql/npgsql/blob/3d4b1d709982468345aca212eceee17e69869087/src/Npgsql/NpgsqlDataReader.cs#L1258

However, I admit it proves your point - Postgres also casts types to the requested type so I'm resolving this conversation.

ptiurin (approved):
Looks ok, as mentioned in Slack `GetSchemaTable` is returning an empty table, which can be misleading.

PR #43, opened 2023-10-15
test: Fir 21501 & 16875 - login + token tests

alexradzin:


FireboltNETSDK/Client/FireboltConnection.cs
```diff
@@ -170,7 +178,7 @@ public FireboltConnection(string connectionString)
         /// Initializes a new instance of <see cref="FireboltConnection"/> with the settings.
         /// </summary>
         /// <param name="stringBuilder">The connection string builder which will be used for building the connection settings.</param>
-        public FireboltConnection(FireboltConnectionStringBuilder stringBuilder)
+        private FireboltConnection(FireboltConnectionStringBuilder stringBuilder)
```
ptiurin, line 181: I don't think this is correct. At least it's not consistent with our readme that suggests instantiating this class directly - https://github.com/firebolt-db/firebolt-net-sdk#creating-and-closing-a-connection

FireboltNETSDK/FireboltClient.cs
```diff
@@ -43,11 +43,11 @@ public class FireboltClient
     private readonly string _jsonContentType = "application/json";
     private readonly string _textContentType = "text/plain";
 
-    public FireboltClient(String id, String secret, String endpoint, String? env, HttpMessageInvoker httpClient)
+    public FireboltClient(String id, String secret, String? endpoint, String? env, HttpMessageInvoker httpClient)
     {
         _httpClient = httpClient;
         _settings = new Lazy<JsonSerializerSettings>(new JsonSerializerSettings());
-        _endpoint = env != null ? $"api.{env}.firebolt.io" : endpoint;
+        _endpoint = env != null ? $"api.{env}.firebolt.io" : endpoint!;
```
ptiurin, line 50: IMO, we should do a check here that either `env` or `endpoint` parameters passed in are not null and raise an exception if it isn't the case. Otherwise with the null forgiving operator we're falsely declaring `_endpoint` to never contain null only to fail somewhere down the line.

ptiurin (commented):
couple of comments

FireboltNETSDK/FireboltClient.cs
alexradzin: As far as I understand endpoint is not needed at all. I planned to remove it in separate PR. What do you think?

FireboltNETSDK/Client/FireboltConnection.cs
alexradzin: FireboltConnection can be instantiated directly using constructor that accepts string, not the builder. The builder should become gory details of the implementation not exposed to the user.

ptiurin: Ok, makes sense.

FireboltNETSDK/FireboltClient.cs
ptiurin: Let's discuss if it's needed on a daily tomorrow. I'd prefer to keep the parameters consistent (as much as we can) across the drivers.
For this PR, let's aim to keep the changes self-contained. If your priorities shift and we never get around to removing it at least let's keep this in a good state. I think an error check for both `env` and `endpoint` not being null will not hurt us or add too much development overhead.

alexradzin: Sire. This was the reason that I have not changed anything here except adding nullability to the endpoint. I can remove it (hopefully). I really do not remember why I added it. In this case nothing will be changed.

ptiurin: That also works, we probably don't want pointless changes bundled in.

alexradzin: Well, I have removed the nullability. It is really not needed.

ptiurin (approved):

PR #45, opened 2023-10-16
refactor: Simplified type conversion; removed redundant code

alexradzin:


FireboltDotNetSdk.Tests/Unit/TypesConverterTest.cs
```diff
@@ -57,9 +58,9 @@ public void ConvertProvidedValues(string columnTypeName, string value, object ex
     public void FailingConvertProvidedValues(string columnTypeName, string value)
     {
         ColumnType columnType = ColumnType.Of(columnTypeName);
-        FireboltException e = Assert.Throws<FireboltException>(() => TypesConverter.ConvertToCSharpVal(value, columnType));
+        FormatException e = Assert.Throws<FormatException>(() => TypesConverter.ConvertToCSharpVal(value, columnType));
         string visualTypeName = char.ToUpper(columnTypeName[0]) + columnTypeName.Substring(1);
-        Assert.That(e.Message, Is.EqualTo($"Error converting {value} to {visualTypeName}"));
+        Assert.That(e.Message, Is.EqualTo($"Input string was not in a correct format."));
```
ptiurin, line 63: I'd argue the previous error message is better, since it's more specific.

ptiurin (approved):
Any noticeable performance difference? Just thinking, while looking at how those functions were named. Let's check by running x10 unit test runs before this change and with it. If the mean does not differ much, we're ok. 
Otherwise lgtm.

FireboltDotNetSdk.Tests/Unit/TypesConverterTest.cs
alexradzin: The problem was that we had nested exceptions; now we do not. There  are cases when nested exceptions help but this is not the case. If parsing fails I do want to deal with it at all. In java we have runtime and checked exceptions. Checked exceptions must be processed somehow; runtime not. Typically parsing exceptions are runtime, so the are thrown as far as it is possible.

PR #50, opened 2023-10-19
feat: implemented FireboltComamnd.ExecuteScalar

alexradzin:


FireboltDotNetSdk.Tests/Integration/ExecuteTest.cs
```diff
@@ -62,6 +62,17 @@ private void executeTest(string connString, string commandText, string? addition
             }
         }
 
+        [TestCase("select 1", 1)]
+        [TestCase("select 'hello'", "hello")]
+        public void ExecuteScalar(string query, object expectedValue)
```
ptiurin, line 67: Let's add another test, where query contains two fields, two rows and verify that `ExecuteScalar` returns only the first row/filed. This way we verify the definition of if from the docs: `Additional columns or rows are ignored.`

alexradzin: Done. Take a look again, please

FireboltDotNetSdk.Tests/Integration/ExecuteTest.cs
```diff
@@ -62,6 +62,21 @@ private void executeTest(string connString, string commandText, string? addition
             }
         }
 
+        [TestCase("select 1", 1)]
+        [TestCase("select 'hello'", "hello")]
+        [TestCase("select 2, 'two'", 2)]
+        [TestCase("select 'three', 3", "three")]
+        [TestCase("select table_schema, table_name from information_schema.tables where table_name = 'tables'", "information_schema")]
+        [TestCase("select table_name, table_schema from information_schema.tables where table_schema = 'information_schema' order by table_schema", "accounts")]
```
ptiurin, line 70: I wouldn't rely on `information_schema`, since we don't control what's in it. Something like `SELECT 'one', 2 UNION ALL SELECT 'two', 3` will give a nice 2x2 result and will almost 100% stay the same.

FireboltNETSDK/Client/FireboltCommand.cs
```diff
@@ -311,7 +311,10 @@ public override int ExecuteNonQuery()
 
         public override object ExecuteScalar()
         {
-            throw new NotImplementedException();
+            using (DbDataReader reader = ExecuteReader())
+            {
+                return reader.Read() && reader.FieldCount > 0 ? reader.GetValue(0) : throw new InvalidOperationException("Cannot get scalar value from empty response");
```
stepansergeevitch, line 316: I believe we should return `Null` for an empty response.
From the docs: https://learn.microsoft.com/en-us/dotnet/api/system.data.common.dbcommand.executescalar?view=net-7.0#remarks
>If the first column of the first row in the result set is not found, a null reference (Nothing in Visual Basic) is returned. If the value in the database is null, the query returns DBNull.Value

ptiurin (approved):
lgtm

PR #53, opened 2023-10-25
feat: FIR-27622 implemented command timeout and cancel()

alexradzin:


FireboltNETSDK/Client/FireboltCommand.cs
```diff
@@ -37,9 +37,12 @@ public class FireboltCommand : DbCommand
     {
         private FireboltConnection? _connection;
         private string? _commandText;
+        private int _commandTimeout = int.MaxValue / 1000;
```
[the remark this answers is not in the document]
alexradzin: It is just easier. default value == max value means that there is no limit. The rest of the implementation is the same. It is easier than managing some kind of if-based logic.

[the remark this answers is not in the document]
ptiurin: And yet the docs state that 0 means no timeout? https://learn.microsoft.com/en-us/dotnet/api/system.data.sqlclient.sqlcommand.commandtimeout?view=dotnet-plat-ext-7.0#remarks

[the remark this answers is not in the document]
alexradzin: Fixed according to the referenced document.

FireboltDotNetSdk.Tests/Integration/ExecuteTest.cs
```diff
@@ -84,6 +84,37 @@ private void executeTest(string connString, string commandText, string? addition
             }
         }
 
+        [Test]
+        public void NoTimeout()
+        {
+            using var conn = new FireboltConnection(SYSTEM_CONNECTION_STRING);
+            conn.Open();
+            DbCommand command = conn.CreateCommand();
+            command.CommandText = "select * from information_schema.tables";
```
ptiurin, line 93: This query makes tests rely on I_S.tables taking a set amount of time. I feel this makes tests brittle wrt changes on the backend and/or test setup (if we have many tables in db they will be slower). I would suggest using something like `"SELECT checksum(*) FROM GENERATE_SERIES(1, 300000000000)"`, where `300000000000` governs time it executes for. Obviously a low value here would suffice.
We use this query throughout to test execution time so we know what to watch out for.

FireboltNETSDK/Client/FireboltCommand.cs
```diff
@@ -166,7 +176,17 @@ public override int CommandTimeout
             }
 
             var database = Connection?.Database != string.Empty ? Connection?.Database : null;
-            Task<string?> t = Connection!.Client.ExecuteQuery(engineUrl, database, Connection?.AccountId, SetParamList, newCommandText);
+
+            CancellationToken token = CancellationToken.None;
+            if (CommandTimeoutMillis != 0)
```
ptiurin, line 181: Doesn't this prevent cancellation of commands without timeout? Why do we have this check?

ptiurin: Yes, I understand. My question is why do we have this `if`. If I understand correctly, it prevents us from cancelling query with no timeout.

alexradzin: Kind of. 0 value of timeout behaviours exactly as "no timeout". If timeout is 0 we pass `CancellationToken.None` to `ExecuteQueryAsync` and does not register event that will cancel this task automatically when timeout is expired. IMHO this is exactly the "no timeout" behaviour.

ptiurin: I understand what the code does, but I don't understand the logic here. If the query has no timeout then I have even more reason to be able to cancel it in case it misbehaves. Timeout and cancellation are not mutually exclusive. User might want to cancel a query from another thread regardless of the timeout set on it.
IMO this not only redundant but harmful.

alexradzin: All fixed. Please review.

FireboltNETSDK/Client/FireboltCommand.cs
```diff
@@ -267,13 +309,8 @@ public void ClearSetList()
...
         }
 
-        /// <summary>
-        /// Not supported. To cancel a command execute it asynchronously with an appropriate cancellation token.
-        /// </summary>
-        /// <exception cref="NotImplementedException">Always throws <see cref="NotImplementedException"/>.</exception>
         public override void Cancel()
         {
-            throw new NotImplementedException();
         }
```
ptiurin, line 314: Shouldn't we keep the old docstring and error? What's the reasoning here?

ptiurin (approved):
Apart from that lgtm.

FireboltNETSDK/Client/FireboltCommand.cs
alexradzin: I've added comment that explains why exception was removed.

PR #54, opened 2023-10-25
feat: FIR-27628 implemented fake transaction

alexradzin:


FireboltDotNetSdk.Tests/Unit/FireboltConnectionTest.cs
```diff
@@ -1,5 +1,7 @@
 ﻿using System.Data;
+using System.Data.Common;
 using System.Net;
+//using System.Transactions;
```
ptiurin, line 4: comment

ptiurin (approved):
Apart from this lgtm

PR #56, opened 2023-11-22
feat: Both identities

alexradzin:


FireboltDotNetSdk.Tests/Integration/IntegrationTest.cs
```diff
@@ -21,13 +22,55 @@ public static string EnvWithDefault(string env_var, string? default_value = null
...
+        protected static string? ClientId = GetEnvironmentVariable("FIREBOLT_CLIENT_ID");
+        protected static string? ClientSecret = GetEnvironmentVariable("FIREBOLT_CLIENT_SECRET");
+        protected static string? UserName = GetEnvironmentVariable("FIREBOLT_USERNAME");
+        protected static string? Password = GetEnvironmentVariable("FIREBOLT_PASSWORD");
+
+        // protected static string ConnectionString()
+        // {
+        //     return ConnectionStringWithout();
+        // }
+
```
ptiurin, line 39: comment

FireboltDotNetSdk.Tests/Integration/SystemEngineTest.cs
```diff
@@ -63,11 +64,19 @@ private void CreateEngine(string engineName, string? spec = null)
             CreateCommand(create_engine_sql).ExecuteNonQuery();
         }
 
-        [TestCase("CREATE DIMENSION TABLE dummy(id INT)")]
+        [TestCase("select 1")]
+        [TestCase("select count(*) from information_schema.tables where table_name = 'tables'")]
+        public void SuccessfulQueryTest(string query)
+        {
+            Assert.That(CreateCommand("select 1").ExecuteScalar(), Is.EqualTo(1));
+        }
```
ptiurin, line 72: `query`'s not used here?

FireboltNETSDK/Client/FireboltConnectionSettings.cs
```diff
@@ -97,5 +97,19 @@ internal FireboltConnectionSettings(FireboltConnectionStringBuilder builder)
...
+            int notNullCount = notNullValues.Count();
+            switch (notNullCount)
+            {
+                case 0: throw new FireboltException($"Either {firstName} or {secondName} parameter is missing in the connection string");
+                case 1: return notNullValues[0];
+                case 2: throw new FireboltException($"Ambiguous values of {firstName} and {secondName}. Use only one of them");
+                // this cannot happen. Added to satisfy compiler and to prevent future bugs.
+                default: throw new InvalidOperationException($"{notNullCount} values for {firstName} or {secondName}. Only one value is legal");
+            }
+        }
```
ptiurin, line 113: I would just split this function into a `GetNotNullValue` and `VerifyInputs`. This way not only the names actually describe what it does but it's also easier to understand what's going on. More visibility on what's happening in the code is better IMO.

FireboltNETSDK/Client/FireboltConnectionState.cs
```diff
@@ -23,17 +23,17 @@ public sealed class FireboltConnectionState
...
 
 
-        public FireboltConnectionState()
-            : this(ConnectionState.Closed, null, 0)
-        {
-        }
+        // public FireboltConnectionState()
+        //     : this(ConnectionState.Closed, null, 0)
+        // {
+        // }
```
ptiurin, line 35: comment

FireboltNETSDK/Client/FireboltConnectionStringBuilder.cs
```diff
@@ -73,16 +99,18 @@ public string ClientSecret
...
         }
 
         /// <summary>
         /// Get the name of the default Account.
         /// </summary>
-        public string Account
+        public string? Account
         {
-            get => GetString(nameof(Account)) ?? throw new FireboltException("Account parameter is missing in the connection string");
+            // TODO: mandatory for v2 + check all parameters!!!
```
ptiurin, line 112: Are those todo's still valid?

FireboltDotNetSdk.Tests/Integration/ConnectionTest.cs
```diff
@@ -14,32 +14,34 @@ internal class ConnectionTest : IntegrationTest
...
         {
-            var connString = $"clientid={ClientId};clientsecret={ClientSecret};account={Account};env={Env}";
-            string expectedDatabase = string.Empty;
-            if (useDatabase)
+            IList<string> withoutParams = new List<string>();
+            if (!useDatabase)
             {
-                connString += $";database={Database}";
-                expectedDatabase = Database;
+                withoutParams.Add(nameof(Database));
```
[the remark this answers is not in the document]
alexradzin: This is indeed list of parameters that should be _ignored_ when creating connection string. To avoid confusion I have renamed this variable and added comment.

FireboltDotNetSdk.Tests/Integration/IntegrationTest.cs
alexradzin: removed

FireboltNETSDK/Client/FireboltConnectionSettings.cs
alexradzin: I have reviewed this function again. I do not know how to split it and avoid code duplication.

FireboltNETSDK/Client/FireboltConnectionStringBuilder.cs
```diff
@@ -145,5 +176,21 @@ public FireboltConnectionSettings BuildSettings()
         {
             return TryGetValue(key, out var value) ? (string)value : null;
         }
+
+        private void InitVersion()
+        {
+            if (Endpoint == null)
```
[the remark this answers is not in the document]
alexradzin: Firebolt provides service. No-one can install his/her own instance of firebolt and connect to it. This is why endpoint is (kind of) obsolete. It was mistake that it was used in FB1, but now it is not needed at all. This is why users should not provide it when they connect to firebolt.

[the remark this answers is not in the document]
stepansergeevitch: Do I understand correctly that endpoint is something like `api.dev.firebolt.io`?
In such case, what will happen if a user, who uses firebolt 1.0 will not provide it? Will it be equal to null?

FireboltNETSDK/Client/FireboltConnectionStringBuilder.cs
```diff
@@ -145,5 +176,21 @@ public FireboltConnectionSettings BuildSettings()
...
+            {
+                Version = 2;
+                return;
+            }
+            // old format
+            if (ClientId != null && ClientSecret != null && UserName == null && Password == null)
```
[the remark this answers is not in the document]
stepansergeevitch: Sorry, misread the code, it's fine here

FireboltNETSDK/Client/FireboltConnectionSettings.cs
ptiurin: This is IMO massively over-engineered function in order to save possibly two lines of duplicate code. It does two things, as I mentioned, "coalesce" firstValue or secondValue (as the name suggests but in an obscure way of counting nulls), it also throws errors which will be confusing to see `GetOneValue` complaining about inputs.

Have one function that verifies the connection string, you can then add as many rules there as you want. This would also be extensible, if we were to add some more constraints or verify that some of the inputs are aphanumeric, etc. So something like `private boolean isConnectionValid(FireboltConnectionStringBuilder builder) throws FireboltExcepion`
For the first non-null value, no need to reinvent the wheel - Apache Commons which we already have as a dependency has `firstNonNull` https://commons.apache.org/proper/commons-lang/javadocs/api-3.1/org/apache/commons/lang3/ObjectUtils.html#firstNonNull%28T...%29

FireboltNETSDK/Client/FireboltConnectionSettings.cs
```diff
@@ -97,5 +99,29 @@ internal FireboltConnectionSettings(FireboltConnectionStringBuilder builder)
...
+            int notNullCount = notNullValues.Count();
+            switch (notNullCount)
+            {
+                case 0: throw new FireboltException($"Either {firstName} or {secondName} parameter is missing in the connection string");
+                case 1: break; // OK
+                case 2: throw new FireboltException($"Ambiguous values of {firstName} and {secondName}. Use only one of them");
+                // this cannot happen. Added to satisfy compiler and to prevent future bugs.
+                default: throw new InvalidOperationException($"{notNullCount} values for {firstName} or {secondName}. Only one value is legal");
+            }
+        }
```
ptiurin, line 125:
```suggestion
        private void ValidateValues(FireboltConnectionStringBuilder builder)
        {
            var userName = builder.UserName;
            var password = builder.Password;
            var clientId = builder.ClientId;
            var clientSecret = builder.ClientSecret;

            if (AreBothProvided(userName, clientId) || AreBothMissing(userName, clientId))
            {
                throw new FireboltException("Configuration error: either username or clientid must be provided but not both");
            }

            if (AreBothProvided(password, clientSecret) || AreBothMissing(password, clientSecret))
            {
                throw new FireboltException("Configuration error: either password or clientsecret must be provided but not both");
            }
        }

        private bool AreBothMissing(string? firstValue, string? secondValue)
        {
            return string.IsNullOrEmpty(firstValue) && string.IsNullOrEmpty(secondValue);
        }

        private bool AreBothProvided(string? firstValue, string? secondValue)
        {
            return !string.IsNullOrEmpty(firstValue) && !string.IsNullOrEmpty(secondValue);
        }
```

FireboltNETSDK/Client/FireboltConnectionSettings.cs
ptiurin: You're right, not sure why I looked at Java. Regardless, it looks ugly to me too, but not because of GetNotNullValue, I like having simple functions that do one thing. The switch there and counting nulls for two values is unnecessary and just contributes to obscurity. Have a look at the suggestion I've added in a separate comment - much simpler, some duplication but it's clear what's happening from the first glance.

FireboltNETSDK/Client/FireboltConnectionStringBuilder.cs
```diff
@@ -33,20 +33,42 @@ public class FireboltConnectionStringBuilder : DbConnectionStringBuilder
         /// Gets or sets the name of the user.
         /// </summary>
         /// <returns>The name of the user.</returns>
-        public string ClientId
+        public string? UserName
         {
-            get => GetString(nameof(ClientId)) ?? throw new FireboltException("ClientId parameter is missing in the connection string");
+            get => GetString(nameof(UserName));// ?? (GetString(nameof(ClientId)) == null ? throw new FireboltException("Either ClientId or UserName must be provided") : null);
+            init => this[nameof(UserName)] = value;
+        }
```
ptiurin, line 40: Here and in 4 other places in this file there's commented code. Did you mean to remove it?

ptiurin (approved):
Please remove commented code or explain why it's there.
Otherwise, lgtm.

PR #58, opened 2023-11-26
feat: added global.json generation

alexradzin:
This is simple but very urgent PR that fixes CI/CD problem introduced by github a couple of weeks ago when they updated default version of .NET that caused our build to fail. File `global.json` generated right after .NET SDK installation fixes this problem.


ptiurin (approved):

ptiurin:
Do we need the same in integration test workflow?

PR #62, opened 2023-12-17
feat: FIR-28769: special treatment of 404 status when accessing account

alexradzin:


ptiurin (commented):
Can we add an integration test for this? We can create an account with restrictive RBAC and another with correct RBAC but not linked to a user and use those in our tests.

alexradzin:
> Can we add an integration test for this? We can create an account with restrictive RBAC and another with correct RBAC but not linked to a user and use those in our tests.

Do we have such tests in other drivers that have the same feature? Let's separate this: first create special account, then implement this test for each driver. BTW I even not sure that this feature is already supported at the server side.

ptiurin:
> Do we have such tests in other drivers that have the same feature? Let's separate this: first create special account, then implement this test for each driver. BTW I even not sure that this feature is already supported at the server side.

We do, all our changes should contain the relevant tests as well. Let's keep the change and tests together in here as well.
A better strategy instead of having one persistent account was discussed on one of the dailies - a test account is created in the test itself, via the SQL and destroyed on the test completion.

alexradzin:
Done

ptiurin (approved):

PR #64, opened 2023-12-28
test: FIR-25166: added long running integration test

alexradzin:


FireboltDotNetSdk.Tests/Integration/ExecuteTest.cs
```diff
@@ -52,11 +52,17 @@ public void ExecuteSelectTest(string commandText)
...
-        public void ExecuteSetTest(string commandText)
+        [Test]
+        [Category("slow")]
+        public void ExecuteLongTest()
         {
-            executeTest(SYSTEM_CONNECTION_STRING, commandText, "SET use_standard_sql=0", true);
+            using var conn = new FireboltConnection(SYSTEM_CONNECTION_STRING);
+            conn.Open();
+            DbCommand command = conn.CreateCommand();
+            command.CommandTimeout = 0; // make command timeout unlimited
```
ptiurin, line 62: I'd still add a large timeout here, just to be safe. This way if the query misbehaves in any way we won't have tests running forever.

ptiurin (approved):
otherwise lgtm

FireboltDotNetSdk.Tests/Integration/ExecuteTest.cs
alexradzin: Good point. I have added timeout.

PR #66, opened 2024-01-24
fix: Keepalive timeout + Fir 29419 release action

stepansergeevitch:
Some tests and ci improvements
- reworked and fixed release action
-  refactored integration tests action to conform to how we have them in other drivers
- fixed keepalive issue and updated integration tests


stepansergeevitch:
Integration tests: https://github.com/firebolt-db/firebolt-net-sdk/actions/runs/7626400253
The failing one is because of a current backend issue with creating service accounts

.github/workflows/release.yml
```diff
@@ -3,15 +3,100 @@ name: Release new version
...
+
+      - name: Version bump
+        run: |
+          sed -i "s/^    <FileVersion>.*<\/FileVersion>/    <FileVersion>${{ steps.tag_generation.outputs.new_version }}<\/FileVersion>/" FireboltNETSDK/FireboltDotNetSdk.csproj
+          sed -i "s/^    <AssemblyVersion>.*<\/AssemblyVersion>/    <AssemblyVersion>${{ steps.tag_generation.outputs.new_version }}<\/AssemblyVersion>/" FireboltNETSDK/FireboltDotNetSdk.csproj
+          sed -i "s/^    <Version>.*<\/Version>/    <Version>${{ steps.tag_generation.outputs.new_version }}<\/Version>/" FireboltNETSDK/FireboltDotNetSdk.csproj
+          git branch
+          git diff
+          git add FireboltNETSDK/FireboltDotNetSdk.csproj
+          git commit -m "Automatic version bump to ${{ inputs.version }}"
```
ptiurin, line 75: I don't think there's `inputs.version` in this wf. Probably `${{ steps.tag_generation.outputs.new_version }}`?

.github/workflows/release.yml
```diff
@@ -3,15 +3,100 @@ name: Release new version
...
+          git push origin ${{ inputs.branch }}
```
ptiurin, line 76: Same here, `inputs.branch` is not defined. We can probably hardcode main, since we're not releasing another branch.

FireboltDotNetSdk.Tests/Integration/ExecuteTest.cs
```diff
@@ -52,17 +53,27 @@ public void ExecuteSelectTest(string commandText)
...
         [Category("slow")]
         public void ExecuteLongTest()
         {
+            var watch = System.Diagnostics.Stopwatch.StartNew();
             using var conn = new FireboltConnection(SYSTEM_CONNECTION_STRING);
             conn.Open();
             DbCommand command = conn.CreateCommand();
             command.CommandTimeout = 0; // make command timeout unlimited
-            command.CommandText = "SELECT checksum(*) FROM generate_series(1, 200000000000)";
+            var max = UserName == null ? 400000000000 : 250000000000;
```
ptiurin, line 65: Let's add a comment that this is a v1/v2 check. I was so confused for a bit why do we check for the UserName.

.github/workflows/release.yml
stepansergeevitch: Right, thanks!

ptiurin (approved):
lgtm

alexradzin (changes requested):
Generally let's avoid mixing-up so irrelevant changes as CI/CD and tests in one PR. In this case if these changes are done in different commits, can you (if it is not difficult) squash your commits into 2 two or even three: for CI,for Keepalive and for unrelated tests.

FireboltDotNetSdk.Tests/Unit/FireboltDataReaderTest.cs
```diff
@@ -616,21 +616,21 @@ public void InfiniteNumberValue()
             string posInf = float.PositiveInfinity.ToString();
             string negInf = float.NegativeInfinity.ToString();
 
-            Assert.That(Assert.Throws<InvalidCastException>(() => reader.GetInt16(0)).Message, Is.EqualTo($"Cannot convert {posInf} to Int16"));
```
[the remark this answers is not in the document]
stepansergeevitch: Just to fix type checks. My IDE complains for example, I don't think it would harm

FireboltDotNetSdk.Tests/Integration/ExecuteTest.cs
```diff
@@ -52,17 +53,28 @@ public void ExecuteSelectTest(string commandText)
...
         public void ExecuteLongTest()
         {
+            var watch = System.Diagnostics.Stopwatch.StartNew();
             using var conn = new FireboltConnection(SYSTEM_CONNECTION_STRING);
             conn.Open();
             DbCommand command = conn.CreateCommand();
             command.CommandTimeout = 0; // make command timeout unlimited
-            command.CommandText = "SELECT checksum(*) FROM generate_series(1, 200000000000)";
+            // Use more rows for FB2.0 since it seems to be faster
+            var max = UserName == null ? 400000000000 : 250000000000;
```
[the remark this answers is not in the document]
stepansergeevitch: Wouldn't that overcomplicate this test?
We would still have this ugly logic and we'll have another abstraction function.
Maybe you have a better vision for it, in such case can you share the example of how you think this should be done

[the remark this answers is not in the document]
alexradzin: IMHO not because this adds some kind of granularity and visibility, also this helps to add support to v3. However it is not the thing I would fight for.

alexradzin (approved):

PR #69, opened 2024-01-30
ci: remove environment parameter from ci

stepansergeevitch:
Removed `environment` parameter from all CI jobs, and set `staging` as a default env

### DIFF
--- a/FireboltDotNetSdk.Tests/Integration/SystemEngineTest.cs
+++ b/FireboltDotNetSdk.Tests/Integration/SystemEngineTest.cs
@@ -234,6 +234,7 @@ namespace FireboltDotNetSdk.Tests
         [Test]
         [Category("v2")]
         [Category("slow")]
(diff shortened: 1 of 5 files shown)


stepansergeevitch:
Integration tests are running, yet failing: https://github.com/firebolt-db/firebolt-net-sdk/actions/runs/7711650196

FireboltDotNetSdk.Tests/Integration/SystemEngineTest.cs
```diff
@@ -233,7 +233,6 @@ private void AssertStartStopEngineAndDropDbTest(Func<FireboltException?, string?
 
         [Test]
         [Category("v2")]
-        [Category("dev")]
```
ptiurin, line 236: does this test work on staging now?

ptiurin (approved):
Just making sure the test works, otherwise lgtm

PR #72, opened 2024-02-01
ci: fix tag generation logic

stepansergeevitch:
Removed character trimming when getting the previous version of a package.


ptiurin (approved):
I think the old code expected tags to contain `v` as a prefix. Let's just remember not to add them from now on.

PR #74, opened 2024-02-05
ci: Fix release build

stepansergeevitch:
A couple of improvements to release action
- Attemt to fix the build by synchronizing build step with how it's done in other actions
- Moved tag pushing to the end of the action

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -68,6 +68,12 @@ jobs:
           git commit -m "Automatic version bump to ${{ steps.tag_generation.outputs.new_version }}"
           git push origin main
         shell: bash
+        
+      - name: Publish tag on github
+        run: |
+          git tag ${{ steps.tag_generation.outputs.new_version }}
+          git push origin ${{ steps.tag_generation.outputs.new_version }}
+        shell: bash
 
       - name: Set up .NET 6.0
         id: dotnet-setup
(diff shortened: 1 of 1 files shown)


ptiurin (commented):
Tag is done before pushing to the software repository in other actions, what are we trying to achieve here? https://github.com/firebolt-db/action-python-release/blob/main/action.yml

alexradzin (approved):

stepansergeevitch:
The problem that has happened is that tag was pushed but release has failed. I believe this logic makes more sense since we should create a tag only after release has succeeded
BTW tag is created after publishing in node sdk as well
https://github.com/firebolt-db/firebolt-node-sdk/blob/main/.github/workflows/release-workflow.yaml#L57

ptiurin:
> I believe this logic makes more sense since we should create a tag only after release has succeeded

This is a chicken and egg problem - if there's a problem with pushing a tag we'll fail, but the driver would already be pushed to the repository. If we rerun it'll be pushed again and I'm not sure what happens if the same tag is pushed twice, will it fail or overwrite? On the other hand, if we fail to push to the remote we have access to quickly delete a tag from our repo and start again.
In any case I'd prefer we keep the release functionality the same across repos, whichever approach we take, and have contingency plans for when it fails.

PR #75, opened 2024-04-25
feat: FIR-30193: improved error handling when JSON parsing fails

alexradzin:


FireboltNETSDK/Client/FireboltCommand.cs
```diff
@@ -330,8 +330,15 @@ private string GetParamValue(object? value)
...
-            var prettyJson = JToken.Parse(Response).ToString(Formatting.Indented);
-            return JsonConvert.DeserializeObject<QueryResult>(prettyJson);
+            try
+            {
+                var prettyJson = JToken.Parse(Response).ToString(Formatting.Indented);
+                return JsonConvert.DeserializeObject<QueryResult>(prettyJson);
+            }
+            catch (JsonReaderException e)
+            {
+                throw new FireboltException($"Failed to execute a query. Invalid response body format. Try again or contact support. {Response}", e);
```
ptiurin, line 340: My only problem here is that if response is quite verbose it will obscure the error. e.g. a large JSON that's missing a } at the end will be printed.
Should we leave the message ending with "Try again or contact support" and perhaps print the response in a debug log? WDYT?

stepansergeevitch: That's the point of the task. We were asked to specifically print the response body when we're not able to parse it

ptiurin: Ok, then I'm asking to re-assess the task. My point still stands, if the response body is large the error will be obscured. I don't think the user will scroll up 100 lines of random json (I'm thinking worst case here) to find the nice error we're outputting. I guess either approach leads to the user contacting support, but one of them will lead them to say "I get a stack trace" another "My query fails on the response body error".
As an option, we can print the error after the response?

alexradzin: Following our discussion I have removed the JSON body from the exception.

ptiurin (approved):

PR #79, opened 2024-05-07
docs: FIR-16886: added docstrings for all standard API methods

alexradzin:


FireboltNETSDK/Client/FireboltConnection.cs
```diff
@@ -149,8 +150,16 @@ public override string ServerVersion
             }
         }
 
+        /// <summary>
+        /// Gets the name of the database to which to connect.
+        /// </summary>
+        /// <returns>The name of the database server to which to connect.</returns>
```
ptiurin, line 156: `database server` implies this is a url, I assume this is the name? Let's remove mention of server altogether to avoid confusion.

FireboltNETSDK/Client/FireboltConnection.cs
```diff
@@ -137,6 +134,10 @@ internal int InfraVersion
         /// <returns>The state of the connection.</returns>
         public override ConnectionState State => _connectionState.State;
 
+        /// <summary>
+        /// Gets database version.
+        /// </summary>
+        /// <returns>The version of the database.</returns>
         public override string ServerVersion
```
ptiurin, line 141: Summary and returns tags are somewhat redundant here (and in some other places).
Can one of them be omitted? What's the best practice here?

ptiurin (approved):
Apart from those two comments, lgtm.

FireboltNETSDK/Client/FireboltConnection.cs
alexradzin: I really do not know about the best practice. You are right that this may look redundant but I saw the same both in java and in C#.

PR #80, opened 2024-05-13
fix: default env value

stepansergeevitch:
Fixed default env value to be `app` rather than `api`


ptiurin (approved):
Do we have a test for this?

stepansergeevitch:
I believe no, and I'm not sure how to do this since we don't want to use prod env for testing

ptiurin:
> I believe no, and I'm not sure how to do this since we don't want to use prod env for testing

I was thinking more of a unit test. Whichever function uses this attribute is going through some if clauses to determine the environment, we can iterate and check these as well as the default.

PR #86, opened 2024-05-22
fix: FIR-33174: parse boolean in old (0/1) and new (false/true) formats

alexradzin:


FireboltNETSDK/Utils/Types.cs
```diff
@@ -80,6 +87,12 @@ public static class TypesConverter
             }
         }
 
+        public static bool ParseBoolean(string str)
+        {
+            bool b;
+            return booleanValues.TryGetValue(str, out b) ? b : throw new FormatException($"String '{str}' was not recognized as a valid Boolean.");
```
ptiurin, line 93: Should this be a case-insensitive comparison? For cases like `True` and `TRUE`?

alexradzin: In fact it should support both `true` and `True` because string value arrives here after calling `ToString()` that  turns `true` into string `True` and `false` into string `False`. 

I agree that this is not obvious, so I can either add this text as a comment or add entries for `True` and `False` to the dictionary. What do you prefer?

ptiurin: I see. Eh, just a comment here is fine so we don't forget.

ptiurin (approved):
lgtm

PR #87, opened 2024-05-22
fix: FIR-33181 re-enabled unit tests

alexradzin:


stepansergeevitch (approved):
I believe this should do the trick, hovewer after investigation I see that a setup fixture runs per namespace, so we might want to splint integration and unit tests into two separate namespaces

PR #88, opened 2024-05-28
feat: FIR-33267: quoted engine and DB name

alexradzin:


FireboltDotNetSdk.Tests/Integration/SystemEngineTest.cs
ptiurin, line 1: I think another test we could add is connecting to a mixed-case engine via a connection string. This way we'll test USE ENGINE/DATABASE that we construct instead of primarily testing the backend.

alexradzin: Added such test ConnectToEngineMixedCase

FireboltDotNetSdk.Tests/Integration/SystemEngineTest.cs
```diff
@@ -284,7 +284,91 @@ public void UseEngineTest()
...
+                string connectionString = ConnectionString(new Tuple<string, string?>[]
+                {
+                    Tuple.Create<string, string?>(nameof(Engine).ToLower(), engineName),
+                    Tuple.Create<string, string?>(nameof(Database).ToLower(), databaseName)
+                });
+                DbConnection connection = new FireboltConnection(connectionString);
+                connection.Open();
+
+                DbCommand command = connection.CreateCommand();
+                command.CommandText = "SELECT 1";
```
ptiurin, line 358: I would do CREATE TABLE test (id INT); INSERT INTO test VALUES (1);
This ensures we're 100% connected to the engine/db and not system engine.

ptiurin (approved):
otherwise lgtm

FireboltDotNetSdk.Tests/Integration/SystemEngineTest.cs
alexradzin: Done

PR #89, opened 2024-05-29
feat: Cache account data

alexradzin:


FireboltNETSDK/FireboltClient2.cs
```diff
@@ -104,11 +106,19 @@ private async Task<T> GetAccountField<T>(string fieldName, string accountName)
...
         await EstablishConnection();
-        // Connecting to system engine by default
-        var result = await GetSystemEngineUrl(_account);
-        if (result.engineUrl != null)
+        string cacheKey = $"{_env}.{_account}";
+        string? systemEngineUrl;
+        systemEngineUrlCache.TryGetValue(cacheKey, out systemEngineUrl);
+        if (systemEngineUrl == null)
+        {
+            // Connecting to system engine by default
```
ptiurin, line 114: nit: Let's move this comment where i makes more sense, otherwise in here it's confusing since even without going into the if block we're still connecting to system engine by default.

FireboltNETSDK/Client/FireboltConnection.cs
```diff
@@ -52,6 +54,8 @@ public FireboltClient Client
         private FireboltClient? _fireboltClient;
         public readonly HashSet<string> SetParamList = new();
         private int _infraVersion = 0;
+        private static IDictionary<string, GetAccountIdByNameResponse> accountCache = new ConcurrentDictionary<string, GetAccountIdByNameResponse>();
+        private ISet<CacheListener> listeners = new HashSet<CacheListener>();
```
ptiurin, line 58: Why do we add listener support here? Any scenario where it would be useful to register internal cache cleanup events?

alexradzin: This is one of the OOD principles: decoupling.

FireboltNETSDK/FireboltClient2.cs
alexradzin: When you added this comment on 2023-07-26 it did not confuse you. 

e3f72f6f (Petro Tiurin     2023-07-26 09:41:47 +0100 257)             // Connecting to system engine by default

Now, the only thing that was changed is an indentation because of the caching. 

Here is the original code:

```
        public override async Task<bool> OpenAsync(CancellationToken cancellationToken)
        {
            if (_fireboltClient == null)
            {
                _fireboltClient = new FireboltClient(ClientId, ClientSecret, Endpoint, Env, HttpClientSingleton.GetInstance());
            }
            await Client.EstablishConnection();
            // Connecting to system engine by default
            var result = await Client.GetSystemEngineUrl(Account);
            EngineUrl = result.engineUrl;
            // Specific engine and database specified
            if (EngineName != null && Database != string.Empty)
            {
                EngineUrl = GetEngineUrlByEngineNameAndDb(EngineName, Database);
                _isSystem = false;
            }
            else if (EngineName != null)
            {
                // If no db provided - try to fetch it
                var database = GetEngineDatabase(EngineName);
                if (database == null)
                {
                    throw new FireboltException($"Engine {EngineName} is attached to a database current user can not access");
                }
                EngineUrl = GetEngineUrlByEngineNameAndDb(EngineName, database);
                _database = database;
                _isSystem = false;
            }
            OnSessionEstablished();
            return EngineUrl != null;
        }



```
Where this comment makes more sense? Please let me know and I will move it.

ptiurin: > When you added this comment on 2023-07-26 it did not confuse you.

Nope, it did not.

Now, if you do a git blame on the current version you'll find out that you in fact changed this section (also indicated by the green background in github). Please re-read the code and the comment above to find a better place for it since it no longer makes sense.

FireboltNETSDK/Client/FireboltConnection.cs
ptiurin: Please elaborate on this. Otherwise OOD principle: overengineering comes to mind.

>  Any scenario where it would be useful to register internal cache cleanup events?

FireboltNETSDK/FireboltClient2.cs
alexradzin: This code was just moved from Connection to FireboltClient2. The only thing that I changed now is adding the cache. It is really not a problem to remove this comment. I really do not know where is the better place for it. Please suggest if you do no agree that the comment that you wrote yourself is misplaced.

FireboltNETSDK/Client/FireboltConnection.cs
alexradzin: I cannot elaborate because I do not understand what's the problem?
I have added cache for account data. I have also added method Cleanup() that clears the cache stored in the connection level. But there is data cached on FireboltClient level. This cache should be cleaned as well. So, I added the following logic. Firebolt client "registers" itself calling method Connection.Regiser(this) (seem FireboltClient's constructor). The method Register adds the client into the list, so than, when Connection.Cleanup() is called it can iterate over this list and clean the clients also. 

Could I store collection of FireboltClients? Yes, surely. But I preferred to define interface CacheListener that so far has only one method (although it can be expanded in future). So, Connection "does not know" that it cleans up FireboltClients. It cleans up cache listeners. This is the decoupling that I mentioned above.

FireboltNETSDK/Client/FireboltConnection.cs
```diff
@@ -114,7 +118,13 @@ internal bool IsSystem
             {
                 if (_accountId == null && Account != null && _isSystem)
                 {
-                    GetAccountIdByNameResponse account = Client.GetAccountIdByNameAsync(Account, CancellationToken.None).GetAwaiter().GetResult();
+                    string cacheKey = $"{Env}.{Account}";
```
stepansergeevitch, line 121: I suggest encapsulating the whole lookup logic into the cache itself. Maybe let's create a separate implementation of the Cache class with this logic included

stepansergeevitch: Same goes for other caches, let's have a generic implementation

stepansergeevitch: We can also have a single registry class for all caches, that will also take case of cleaning them up simultaneousy, which will allow us to simplify the listeners logic

stepansergeevitch: See node sdk example for similar idea: https://github.com/firebolt-db/firebolt-node-sdk/commit/5bf24b1a6a2fa5199e7320de08892b3a1657b812

FireboltNETSDK/FireboltClient2.cs
ptiurin: pushed this [change](https://github.com/firebolt-db/firebolt-net-sdk/pull/89/commits/36c8cd5a658910cd99b1b95eee4ec469581e35ca) to your branch, please remember to sync.
As I said, you've added an if around it. It makes sense to move the comment out since we're either way (in and out of if) connecting to system engine (or remove it altogether). 
It does not matter who wrote the code, you touch it - you own it, as simple as that. I really don't understand how this small nitpick has become such an issue.

FireboltNETSDK/Client/FireboltConnection.cs
ptiurin: > I cannot elaborate because I do not understand what's the problem?

There is no problem. I was asking you to clarify the use case for it. This comment is a _question_.
As per my (limited) understanding we were doing a cleanup anyway so I did not see where the listener was used. All I needed is an explanation

ptiurin:
Let's add a test that verifies the cache has been used. Possibly via counting the number of http requests made to account id endpoint.

FireboltNETSDK/Client/FireboltConnection.cs
alexradzin: Anyway this is irrelevant now because this code is removed after discussion with Stepan

FireboltNETSDK/Client/FireboltConnection.cs
alexradzin: We spent 1 hour discussing the issue and hopefully now it is OK from Stepan's point of view

alexradzin:
> Let's add a test that verifies the cache has been used. Possibly via counting the number of http requests made to account id endpoint.


SuccessfulLoginTwice - the new test that does it

ptiurin (approved):
lgtm

PR #91, opened 2024-06-04
feat: FIR-33439: added FireboltClientFactory

alexradzin:


FireboltNETSDK/Client/FireboltClientFactory.cs
```diff
@@ -0,0 +1,55 @@
+using System.Data.Common;
+using System.Security.Permissions;
+
+namespace FireboltDotNetSdk.Client
+{
+    public sealed class FireboltClientFactory : DbProviderFactory
```
stepansergeevitch, line 6: nit: I don't think it's related to a Client, maybe better to name it FireboltProviderFactory or something?

stepansergeevitch (approved):

FireboltNETSDK/Client/FireboltClientFactory.cs
alexradzin: This is what I did in the beginning but then renamed it following the MS SQL naming convention: https://github.com/microsoft/referencesource/blob/master/System.Data/System/Data/SqlClient/SqlClientFactory.cs

The same is about MySQL:

https://github.com/mysql/mysql-connector-net/blob/87939c99fdad68b30803e9f4bb115e17cfabb8ce/MySQL.Data/src/MySqlClientFactory.cs

PR #92, opened 2024-06-09
feat: FIR-33601: use catalogs table instead of databases table if exists

alexradzin:


FireboltDotNetSdk.Tests/Unit/FireboltConnectionTest.cs
```diff
@@ -509,9 +509,11 @@ public void SuccessfulLoginTwice()
...
 
 
-        [TestCase("Running", "api.firebolt.io", "api.firebolt.io")]
-        [TestCase("ENGINE_STATE_RUNNING", "api.firebolt.io?account_id=01hf9pchg0mnrd2g3hypm1dea4&engine=max_test", "api.firebolt.io")]
-        public void SuccessfulLoginWithEngineName(string engineStatus, string engineUrl, string expectedEngineUrl)
+        [TestCase("Running", "api.firebolt.io", "", "api.firebolt.io")] // no catalogs table
+        [TestCase("ENGINE_STATE_RUNNING", "api.firebolt.io?account_id=01hf9pchg0mnrd2g3hypm1dea4&engine=max_test", "", "api.firebolt.io")] // no catalogs table
+        [TestCase("Running", "api.firebolt.io", "[\"catalogs\"]", "api.firebolt.io")] // catalogs table exists
+        [TestCase("ENGINE_STATE_RUNNING", "api.firebolt.io?account_id=01hf9pchg0mnrd2g3hypm1dea4&engine=max_test", "[\"catalogs\"]", "api.firebolt.io")] // catalogs table exists
+        public void SuccessfulLoginWithEngineName(string engineStatus, string engineUrl, string catalogs, string expectedEngineUrl)
```
ptiurin, line 516: My only problem here is that we don't verify that when catalogs are available we query catalogs and not databases. Can we verify that when catalogs is present we don't query `information_schema.databases` and vice versa?

alexradzin: This is a good idea but it is not so simple due to HttpRequestMessage is not a simple object. When I try to access at the end I get exception that the object is already disposable. It seems that it is very tricky to implement this idea although theoretically it is a good one.

ptiurin: Let's try to find a solution, since otherwise this feature is essentially is untested.

alexradzin: I tried. I spent 2 hours. You are welcome to try more.

ptiurin: Tried more, have a look here: https://github.com/firebolt-db/firebolt-net-sdk/pull/94
It's by no means the most elegant solution, but it confirms we can test this condition.
The main trick here is to make sure the condition only succeeds for one case ( I had failures when trying to check that "catalogs" is in query, since it matches two queries - one for checking if this table exists and another one when actually querying I_S.catalogs) since it returns the same object each time and if it's consumed once it will be destroyed.

ptiurin (approved):

PR #95, opened 2024-07-05
feat(FIR-33629): Supporting structured errors

ptiurin:
Changing other tests in order to be future-proof when JSON body error becomes the norm. We're only concerned that FireboltException or its subclasses are returned.


stepansergeevitch (approved):
LGTM, please address failing code checks

PR #96, opened 2024-07-10
ci: Cleanup and standardise our workflows

ptiurin:
Removing redundant and obsolete parameters from our integration test workflows. Keeping it in line with other repos.

Workflows are working, but some v1 tests are failing. Will make the fixes separately for those.

### DIFF
--- a/.github/workflows/integration-tests-v1.yml
+++ b/.github/workflows/integration-tests-v1.yml
@@ -33,8 +33,6 @@ jobs:
     steps:
       - name: Check out code
         uses: actions/checkout@v2
-        with:
-          ref: ${{ inputs.branch }}
 
       - name: Set up .NET 6.0
         id: dotnet-setup
@@ -49,9 +47,6 @@ jobs:
       - name: Install dependencies
         run: |
           dotnet restore
-
-      - name: Keep environment name in the summary
-        run: echo '### Ran integration tests against ${{ inputs.environment }} ' >> $GITHUB_STEP_SUMMARY
 
       - name: Setup database and engine
         if: ${{ inputs.database == '' }}
--- a/.github/workflows/integration-tests-v1.yml
+++ b/.github/workflows/integration-tests-v1.yml
@@ -3,21 +3,21 @@ on:
   workflow_dispatch:
     inputs:
       database:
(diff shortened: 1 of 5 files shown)


.github/workflows/integration-tests-v1.yml
```diff
@@ -33,8 +33,6 @@ jobs:
     steps:
       - name: Check out code
         uses: actions/checkout@v2
-        with:
-          ref: ${{ inputs.branch }}
```
stepansergeevitch, line 37: will this still checkout a correct branch?

.github/workflows/integration-tests-v2.yml
```diff
@@ -118,4 +103,4 @@ jobs:
           FIREBOLT_CLIENT_ID: ${{ secrets.FIREBOLT_CLIENT_ID_STG_NEW_IDN  }}
           FIREBOLT_CLIENT_SECRET: ${{ secrets.FIREBOLT_CLIENT_SECRET_STG_NEW_IDN  }}
         run: |
-          dotnet test --filter '(Category=Integration|Category=IntegrationSetup)&(Category=${{ inputs.category }}|Category=general)&Name!=ExecuteLongTest' -l "console;verbosity=normal"
+          dotnet test --filter '(Category=Integration|Category=IntegrationSetup)&(Category=engine-v2|Category=general)&Name!=ExecuteLongTest' -l "console;verbosity=normal"
```
stepansergeevitch, line 106: should we run both `v2` and `engine-v2` here?

.github/workflows/integration-tests-v1.yml
ptiurin: It will now checkout the correct branch. Before `inputs.branch` was not defined at all.
When ref is omitted this defaults to the branch that triggered the workflow.

.github/workflows/integration-tests-v2.yml
ptiurin: I think tests that are compatible with engines v2 are already marked as `engine-v2` and those that don't are marked only `v2`. Some tests have both categories so the line in question will run them.

stepansergeevitch (approved):

PR #97, opened 2024-07-10
test: Remove obsolete tests

ptiurin:
System engine is no longer supported in v1 so I'm removing tests that are now broken because of it.
I'm also removing purely v2 tests. e.g. ATTACH ENGINE is not a valid command in engines v2.


ptiurin:
Test run successful: https://github.com/firebolt-db/firebolt-net-sdk/actions/runs/9875474361

stepansergeevitch (approved):

PR #106, opened 2024-09-23
fix: FIR-36807 command timeout parameter is not working in net

stepansergeevitch:
Fixed async execution functions to account for the command timeout value, that's set for a command

### DIFF
--- a/FireboltDotNetSdk.Tests/Integration/ExecuteTest.cs
+++ b/FireboltDotNetSdk.Tests/Integration/ExecuteTest.cs
@@ -153,16 +153,6 @@ namespace FireboltDotNetSdk.Tests
             }
         }
 
-        [Timeout(20)]
-        public void TimeoutThrowsError()
-        {
-            using var conn = new FireboltConnection(SYSTEM_CONNECTION_STRING);
-            conn.Open();
-            DbCommand command = conn.CreateCommand();
-            command.CommandText = LONG_QUERY;
-            command.CommandTimeout = 10;
-        }
-
         [Test]
         [Category("general")]
         public void ShortTimeout()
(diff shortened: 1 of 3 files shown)


FireboltDotNetSdk.Tests/Integration/ExecuteTest.cs
```diff
@@ -121,7 +123,44 @@ public void WithTimeout(int? timeout)
...
+
+        [Timeout(20)]
+        public void TimeoutThrowsError()
+        {
+            using var conn = new FireboltConnection(SYSTEM_CONNECTION_STRING);
+            conn.Open();
+            DbCommand command = conn.CreateCommand();
+            command.CommandText = LONG_QUERY;
+            command.CommandTimeout = 10;
         }
```
ptiurin, line 164: Is this used anywhere?

FireboltNETSDK/Client/FireboltCommand.cs
```diff
@@ -178,15 +178,19 @@ private DbDataReader CreateDbDataReader(QueryResult? queryResult)
             {
                 commandText = ValidateSetCommand(commandText.Remove(0, 4).Trim());
                 SetParamList.Add(commandText);
-                try
+                if (Connection != null)
```
ptiurin, line 181: Do we need this check? We already verify Connection is not null in https://github.com/firebolt-db/firebolt-net-sdk/pull/106/files#diff-9f5e4d8a0218e3f437d1707871b746709b804521e38023b2328a94ffb378452aR168

ptiurin (approved):
Couple of comments, otherwise looks good

FireboltNETSDK/Client/FireboltCommand.cs
stepansergeevitch: Good point, I did this for type checker

FireboltDotNetSdk.Tests/Integration/ExecuteTest.cs
stepansergeevitch: No, this is a leftover

PR #110, opened 2024-11-15
fix: Use await in FireboltClient2.cs private methods. Don't change st…

el1b0t:
When using FireboltConnection.OpenAsync, synchronous calls were being made which would then in turn cause blocking behavior. I updated the version to 1.3.5 because I needed a new version of the nupkg file for testing, so I apologize if that was incorrect.

Also, in order to get the project to build, I had to fix a couple of exceptions that were being re-thrown so as not to alter their stack traces.

### DIFF
--- a/FireboltNETSDK/Client/FireboltCommand.cs
+++ b/FireboltNETSDK/Client/FireboltCommand.cs
@@ -203,10 +203,10 @@ namespace FireboltDotNetSdk.Client
                 {
                     await Connection.ValidateConnection(cancellationToken);
                 }
-                catch (AggregateException e)
+                catch (AggregateException)
                 {
                     SetParamList.Remove(commandText);
-                    throw e;
+                    throw;
                 }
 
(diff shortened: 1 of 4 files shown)


stepansergeevitch (approved):
Hey @el1b0t 
Thank a lot for your contribution!
Please fix the linter issue before merging this
> /home/runner/work/firebolt-net-sdk/firebolt-net-sdk/FireboltNETSDK/FireboltClient2.cs(221,1): error WHITESPACE: Fix whitespace formatting. Replace 9 characters with '\n\s\s\s\s'. [/home/runner/work/firebolt-net-sdk/firebolt-net-sdk/FireboltNETSDK/FireboltDotNetSdk.csproj]

el1b0t:
Thanks for the merge, @stepansergeevitch ! Will there be a 1.3.5 release with this code in place, on nuget.org, in the near future?

PR #113, opened 2024-12-04
feat: Fir 38077 ecosystem support for geography type for net

stepansergeevitch:
Added support for geography type


ptiurin (approved):
lgtm

PR #114, opened 2024-12-18
feat: Fir 38129 ecosystem support for struct type for net

stepansergeevitch:
Added support for struct type


FireboltDotNetSdk.Tests/Unit/ColumnTypeTest.cs
```diff
@@ -105,6 +133,122 @@ public void CreateColumnTypeWithProvidedColumnTypeNamesTest(String columnTypeNam
...
+    {
+        String columnTypeName = "struct(a int, s struct(a array(int), b text null))";
+        
+        ColumnType columnType = ColumnType.Of(columnTypeName);
+        if (columnType is StructType structType)
+        Assert.That(columnType.Type, Is.EqualTo(expectedType));
+        Assert.That(columnType.Nullable, Is.EqualTo(expectedIsNullable));
+        Assert.Null(columnType.Precision);
+        Assert.Null(columnType.Scale);
+    }*/
```
ptiurin, line 250: should this be removed?

FireboltDotNetSdk.Tests/Unit/TypesConverterTest.cs
```diff
@@ -140,6 +151,60 @@ public void ConvertInvalidValue()
...
+        Assert.That(result, Is.EqualTo(expected));
+    }
+
+    public static IEnumerable<TestCaseData> StructTestCase()
+    {
+        yield return new TestCaseData(
+            new Dictionary<String, object> { { "a", 1 }, { "b", "value" } },
+            new Dictionary<String, object> { { "a", 1 }, { "b", "value" } },
+            "struct(a int, b string)"
+        );
```
ptiurin, line 177: Let's add some nulls and columns with spaces, just to check that converting value works in those cases.

ptiurin (commented):
Also, do we have any integration tests for this?

FireboltDotNetSdk.Tests/Unit/ColumnTypeTest.cs
```diff
@@ -10,27 +10,41 @@ public void CreateColumnTypeWithArrayTypeTest()
...
         ColumnType columnType = ColumnType.Of(type);
-        Assert.That(columnType.Type, Is.EqualTo(FireboltDataType.Array));
-        Assert.True(columnType.Nullable);
-        Assert.Null(columnType.Precision);
-        Assert.Null(columnType.Scale);
-        Assert.NotNull(columnType.InnerType);
-        Assert.True(columnType.InnerType!.Nullable);
-        Assert.That(columnType.InnerType.Type, Is.EqualTo(FireboltDataType.Int));
+        if (columnType is ArrayType arrayType)
+        {
```
ptiurin, line 35: Do we need everything wrapped in if here? It makes it a bit harder to read than it was before and conditionals in tests are a slippery slope.
Would this work?
```
if not(columnType is ArrayType arrayType) {
            Assert.Fail("ArrayType expected");
}
...
```
This way we keep it like an assert.

ptiurin (approved):
otherwise lgtm

FireboltDotNetSdk.Tests/Unit/ColumnTypeTest.cs
stepansergeevitch: But then we would still need an else clause, since the scope of `arrayType` is this if statement

PR #118, opened 2025-04-25
fix: FIR-45494 Incorrect array typing in FireboltDataReader

bogdantruta-firebolt:
I changed the array casting to use the .NET Type that can be deducted from the response information, so we have less casting(no need to go into each array element to cast it to a specific type) and also we now return an array of an appropriate Type, rather than returning object[].
Also changed tests to use the expectedType since now we cast correctly for arrays.

### DIFF
--- a/FireboltDotNetSdk.Tests/Integration/ExecuteTest.cs
+++ b/FireboltDotNetSdk.Tests/Integration/ExecuteTest.cs
@@ -58,7 +58,6 @@ namespace FireboltDotNetSdk.Tests
         }
 
         [Test, Timeout(2400000)]
-        [Category("general")]
         [Category("slow")]
         public void ExecuteLongTest()
         {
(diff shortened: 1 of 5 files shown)


ptiurin (approved):
apart from that lgtm

FireboltDotNetSdk.Tests/Integration/ExecuteTest.cs
```diff
@@ -494,25 +493,57 @@ public void ExecuteServiceAccountLoginWithMissingSecret(string secretField)
...
-            ExecuteTest(SYSTEM_CONNECTION_STRING, "select [1,NULL,3]", new int?[] { 1, null, 3 }, typeof(int[]));
+            ExecuteTest(SYSTEM_CONNECTION_STRING, "select [1,3]", new int?[] { 1, 3 }, typeof(int[]));
         }
 
         [Test]
         [Category("general")]
         public void ExecuteSelectTwoDimensionalArray()
         {
-            ExecuteTest(SYSTEM_CONNECTION_STRING, "select [[1,NULL,3]]", new int?[1][] { new int?[] { 1, null, 3 } }, typeof(int[][]));
+            ExecuteTest(SYSTEM_CONNECTION_STRING, "select [[1,3]]", new int?[1][] { new int?[] { 1, 3 } }, typeof(int[][]));
```
stepansergeevitch, line 521: Why did you remove NULL here?

stepansergeevitch (approved):
LGTM, please resolve existing comments

FireboltDotNetSdk.Tests/Integration/ExecuteTest.cs
bogdantruta-firebolt: I created other tests with nullable values that are basically this test, and then added some not nullable tests as well (git treated this one as the new not nullable one due to the name remaining the same)

bogdantruta-firebolt:
Integration test run: https://github.com/firebolt-db/firebolt-net-sdk/actions/runs/14707820433

PR #120, opened 2025-05-15
fix: FIR-45674 Improve isDBNull method

bogdantruta-firebolt:
As a client suggestion, we improved the isDBNull method in the cases the value is not null by skipping value converting from the GetValueSafely method.


ptiurin (commented):
Looks good, do we have tests for this?

bogdantruta-firebolt:
> Looks good, do we have tests for this?

The previous tests that were focused on this, getValue and other reader methods

PR #122, opened 2025-05-26
refactor: Prepare streaming implementation

bogdantruta-firebolt:
In order to implement streaming in .NET, I needed to use the HttpClient for a call option that would not buffer the result in memory. In this way, I also revised the logic in the client.

SendAsync should only throw an error or return a response, so the nullable generic was removed for simplicity. In this way, I removed the null response tests.

I also separated some logic into separate methods for reducing complexity and making things easier for maintenance.

All other changes are Sonar proposed.

### DIFF
--- a/FireboltNETSDK/FireboltClient.cs
+++ b/FireboltNETSDK/FireboltClient.cs
@@ -134,8 +134,7 @@ public abstract class FireboltClient
         if (setParams.Length > 0)
             queryStr.Append('&').Append(setParams);
         urlBuilder.Query = queryStr.ToString();
-        var url = urlBuilder.Uri.ToString();
-        return url;
+        return urlBuilder.Uri.ToString();
     }
 
     private StringBuilder GetQueryString(string? databaseName, string? accountId, string outputFormat = "JSON_Compact")
@@ -149,11 +148,11 @@ public abstract class FireboltClient
         {
             parameters["account_id"] = accountId;
         }
-        if (!_connection.IsSystem && _connection.InfraVersion >= 2 && _connection.EngineName != null)
+        if (_connection is { IsSystem: false, InfraVersion: >= 2, EngineName: not null })
         {
             parameters["engine"] = _connection.EngineName;
         }
-        if (!_connection.IsSystem && _connection.InfraVersion >= 2)
+        if (_connection is { IsSystem: false, InfraVersion: >= 2 })
         {
             parameters["query_label"] = Guid.NewGuid().ToString();
         }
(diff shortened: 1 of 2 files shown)


FireboltNETSDK/FireboltClient1.cs
```diff
@@ -36,13 +36,13 @@ public class FireboltClient1 : FireboltClient
...
     };
-    public FireboltClient1(FireboltConnection connection, string id, string secret, string endpoint, string? env, string account, HttpMessageInvoker httpClient) : base(connection, id, secret, endpoint, env, null, httpClient)
+    public FireboltClient1(FireboltConnection connection, string id, string secret, string endpoint, string? env, string account, HttpClient httpClient) : base(connection, id, secret, endpoint, env, null, httpClient)
     {
         _settings = new Lazy<JsonSerializerSettings>(new JsonSerializerSettings());
         _account = account;
     }
 
-    public override async Task<ConnectionResponse> ConnectAsync(string? engineName, string databaseName, CancellationToken cancellationToken)
+    public override async Task<ConnectionResponse> ConnectAsync(string? engineName, string database, CancellationToken cancellationToken)
```
ptiurin, line 45: Why are we renaming this? Looks consistent with `engineName` for me.

FireboltNETSDK/FireboltClient1.cs
```diff
@@ -70,26 +70,26 @@ public override async Task<ConnectionResponse> ConnectAsync(string? engineName,
...
     }
 
-    protected override Task<LoginResponse> Login(string username, string password, string env)
+    protected override Task<LoginResponse> Login(string id, string secret, string env)
     {
-        var credentials = new UsernamePasswordLoginRequest(username, password);
-        return SendAsync<LoginResponse>(HttpMethod.Post, $"https://{_endpoint}{AUTH_USERNAME_PASSWORD_URL}", JsonConvert.SerializeObject(credentials, _settings.Value), _jsonContentType, false, CancellationToken.None, false);
+        var credentials = new UsernamePasswordLoginRequest(id, secret);
+        return SendAsync<LoginResponse>(HttpMethod.Post, $"https://{_endpoint}{AUTH_USERNAME_PASSWORD_URL}", JsonConvert.SerializeObject(credentials, _settings.Value), _jsonContentType, false, false, CancellationToken.None);
     }
```
ptiurin, line 93: Firebolt 1.0 used username and password so I believe these make sense here.

FireboltNETSDK/FireboltClient.cs
```diff
@@ -109,21 +109,38 @@ public FireboltClient(FireboltConnection connection, string id, string secret, s
...
             Port = -1
         };
 
-        var parameters = new Dictionary<string, string>() { { "output_format", "JSON_Compact" } };
+        var queryStr = GetQueryString(databaseName, accountId);
+        if (setParams.Length > 0)
+            queryStr.Append('&').Append(setParams);
+        urlBuilder.Query = queryStr.ToString();
+        var url = urlBuilder.Uri.ToString();
+        return url;
```
ptiurin, line 138: nit: `return urlBuilder.Uri.ToString();`?

ptiurin (commented):
otherwise lgtm

FireboltNETSDK/FireboltClient1.cs
ptiurin: I would avoid touching 1.0 unless we need to. fingers crossed it will be deprecated soon enough.

ptiurin (approved):

PR #124, opened 2025-05-27
feat: FIR-43726 add streaming to net sdk

bogdantruta-firebolt:
Implemented streaming in .NET

Also did a fix where I used to dismiss error checking if http response code == 200, but this is not right if the server would get error while processing a query. This is also true for JSON_Compact output format and can be tested with this query:
`select 1/(i-100000) as a from generate_series(1,100000) as i`

### DIFF
--- a/FireboltDotNetSdk.Tests/Integration/StreamingQueryTest.cs
+++ b/FireboltDotNetSdk.Tests/Integration/StreamingQueryTest.cs
@@ -226,7 +226,6 @@ namespace FireboltDotNetSdk.Tests
                 Assert.That(e.Message, Does.Contain("Line 1, Column 9: Division by zero\n" +
                                                              "select 1/(i-100000) as a from generate_series(1,...\n"));
             }
-
             Assert.Multiple(() =>
             {
                 Assert.That(oneReadAtLeast, Is.True, "Expected at least one row to be read before the exception was thrown.");
--- a/FireboltNETSDK/Client/FireboltStreamingDataReader.cs
+++ b/FireboltNETSDK/Client/FireboltStreamingDataReader.cs
@@ -182,10 +182,17 @@ namespace FireboltDotNetSdk.Client
         private void HandleError(string line)
         {
             _endOfStream = true;
-            var json = JsonConvert.DeserializeObject<StreamingJsonFinishError>(line);
+            StreamingJsonFinishError? json;
+            try
+            {
+                json = JsonConvert.DeserializeObject<StreamingJsonFinishError>(line);
+            }
+            catch (System.Exception e)
+            {
+                throw new FireboltException("Failed to parse JSON from stream on line: " + line, e);
+            }
             if (json?.MessageType == FinishWithErrors)
                 throw new FireboltStructuredException(json.Errors);
-
             throw new FireboltException("Failed to parse JSON from stream. Unexpected messageType: " + json?.MessageType + " in line: " + line);
         }
 


FireboltNETSDK/Client/FireboltCommand.cs
```diff
@@ -268,6 +268,25 @@ private string GetServerSideQueryParameters()
             );
         }
 
+        private async Task<StreamReader> ExecuteStreamingCommandAsync(string commandText, CancellationToken cancellationToken = default)
```
stepansergeevitch, line 271: I see a couple of issues with this method
1. It duplicates a lot of code from `ExecuteCommandAsync`
2. It doesn't respect the `CommandTimeoutMillis` property. For `ExecuteCommandAsync`, we wrap it with `ExecuteCommandAsyncWithCommandTimeout`, which adds timeout handling. I believe we should do the same here

FireboltNETSDK/Client/FireboltStreamingDataReader.cs
```diff
@@ -0,0 +1,197 @@
...
+                    break;
+            }
+
+            return false;
+        }
+
+        private void InitializeMetadata()
+        {
+            var line = _streamReader.ReadLine();
+            if (string.IsNullOrWhiteSpace(line)) throw new FireboltException("Failed to read line from stream");
```
stepansergeevitch, line 166: Should we also set `_endOfStream=true` here to make sure no further (potentially corrupted) data is read?

FireboltNETSDK/Client/FireboltStreamingDataReader.cs
```diff
@@ -0,0 +1,197 @@
...
+
+            var line = _streamReader.ReadLine();
+            if (string.IsNullOrWhiteSpace(line)) throw new FireboltException("Failed to read line from stream");
+
+            var json = JsonConvert.DeserializeObject<StreamingJsonData>(line);
+            var messageType = json?.MessageType;
+
+            switch (messageType)
+            {
+                case "DATA":
```
stepansergeevitch, line 147: Let's put `START`, `DATA`, `FINISH_SUCCESSFULLY` and `FINISH_WITH_ERRORS` into constants.

FireboltNETSDK/Client/FireboltStreamingDataReader.cs
```diff
@@ -0,0 +1,197 @@
...
+                    json?.Data.ForEach(_currentRowQueue.Enqueue);
+                    return true;
+
+                case "FINISH_SUCCESSFULLY":
+                    _endOfStream = true;
+                    break;
+
+                default:
```
stepansergeevitch, line 155: let's update it to `case FINISH_WITH_ERRORS: HandleError(...) default: throw Exception("unexpected message type ...")`

FireboltNETSDK/Client/FireboltStreamingDataReader.cs
```diff
@@ -0,0 +1,197 @@
...
+        }
+
+        private void InitializeMetadata()
+        {
+            var line = _streamReader.ReadLine();
+            if (string.IsNullOrWhiteSpace(line)) throw new FireboltException("Failed to read line from stream");
+
+            var json = JsonConvert.DeserializeObject<StreamingJsonStart>(line);
+
+            if (json?.MessageType != "START") HandleError(line);
```
stepansergeevitch, line 170: Here let's also be more verbose and mention that we've received an unexpected message type

stepansergeevitch: Since right now we would just throw a generic `Failed to parse JSON from stream:`

FireboltDotNetSdk.Tests/Integration/StreamingQueryTest.cs
```diff
@@ -0,0 +1,277 @@
...
+            using var reader = command.ExecuteStreamedQuery();
+            Assert.Multiple(() =>
+            {
+                Assert.That(reader.Read(), Is.EqualTo(true));
+                Assert.That(reader.GetInt32(0), Is.EqualTo(1));
+                Assert.That(reader.Read(), Is.EqualTo(true));
+                Assert.That(reader.GetInt32(0), Is.EqualTo(1));
+                Assert.That(reader.Read(), Is.EqualTo(false), "Expected no more rows to be read after the second row.");
+            });
+        }
```
stepansergeevitch, line 275: Please also add a test in which we make sure we receive more than one `DATA` record. See [go sdk example](https://github.com/firebolt-db/firebolt-go-sdk/blob/main/connection_common_integration_test.go#L448-L449)

FireboltNETSDK/Client/FireboltStreamingDataReader.cs
bogdantruta-firebolt: sure, why not, I was thinking it shouldn't be the case since it should be used as a resource inside try/catch, but it wouldn't hurt to be sure

FireboltDotNetSdk.Tests/Integration/StreamingQueryTest.cs
bogdantruta-firebolt: the first test with this quert `select 1 from generate_series(1, 2500000)` returns an approximately 1gb response, so there is such a test

FireboltNETSDK/Client/FireboltStreamingDataReader.cs
```diff
@@ -0,0 +1,205 @@
...
+            var json = JsonConvert.DeserializeObject<StreamingJsonStart>(line);
+
+            if (json?.MessageType != Start) HandleError(line);
+            _metas = json?.Meta ?? new List<Meta>();
+        }
+
+        private void HandleError(string line)
+        {
+            _endOfStream = true;
+            var json = JsonConvert.DeserializeObject<StreamingJsonFinishError>(line);
```
stepansergeevitch, line 185: Here I believe we can also receive an error during JSON parsing. Let's also wrap it with a meaningful error message

stepansergeevitch (approved):
LGTM, one comment

PR #128, opened 2025-08-18
feat: FIR-46461 support remove parameters header

bogdantruta-firebolt:


FireboltNETSDK/FireboltClient.cs
```diff
@@ -454,7 +456,7 @@ protected bool ProcessParameters(FireboltConnectionStringBuilder connectionBuild
                         _connection.AccountId = kv[1];
                         break;
                     default:
-                        if (!kv[1].Equals(_queryParameters[kv[0]]))
+                        if (!_queryParameters.ContainsKey(kv[0]) || !kv[1].Equals(_queryParameters[kv[0]]))
```
stepansergeevitch, line 459: Why do we need this check? I think we should update the parameter if it's overriding the existing value

PR #129, opened 2025-08-19
ci: fix release script

bogdantruta-firebolt:


.github/workflows/release.yml
ptiurin, line 1: We can remove `Checkout release action repo` step since we're using a different action

ptiurin: Also python setup and python dependencies

PR #132, opened 2025-08-20
fix: version fix and correct tag output name

bogdantruta-firebolt:


goprean (approved):
is 1.7.1 the version that were are releasing?

PR #135, opened 2025-09-26
feat: FIR-48856 add implicit struct deserialization to custom objects with mapping

bogdantruta-firebolt:
This PR adds support for deserializing struct directly into .NET classes, using the default snake case convention based on property names, or by using the new FireboltStructNameAttribute to map properties with a different name than the one from the result.

### DIFF
--- a/FireboltDotNetSdk.Tests/Integration/DataTypesTest.cs
+++ b/FireboltDotNetSdk.Tests/Integration/DataTypesTest.cs
@@ -2,7 +2,10 @@ using System.Data;
 using System.Data.Common;
 using System.Text;
 using FireboltDotNetSdk.Client;
+using FireboltDotNetSdk.Utils;
 using static System.DateTime;
+using System.Reflection;
+using System.Text.RegularExpressions;
 
 namespace FireboltDotNetSdk.Tests
 {
@@ -69,7 +72,7 @@ namespace FireboltDotNetSdk.Tests
                 "        ['2019-07-31 01:01:01',null]::array(timestamp)                                   as col_timestamp_null_array,\n" +
                 "        null::array(timestamp)                                                           as col_timestamp_array_null,\n" +
                 "        [['2019-07-31 01:01:01','2019-08-01 00:00:00'],null]::array(array(timestamp))    as col_timestamp_array_null_array,\n" +
-                "        [['2019-07-31 01:01:01','2019-08-01 00:00:00'],[]]::array(array(timestamp))      as col_timestamp_array_null_array,\n" +
+                "        [['2019-07-31 01:01:01','2019-08-01 00:00:00'],[]]::array(array(timestamp))      as col_timestamp_array_array,\n" +
                 "        [['2019-07-31 01:01:01',null],null]::array(array(timestamp))                     as col_timestamp_null_array_null_array,\n" +
                 "        null::array(array(timestamp))                                                    as col_timestamp_array_array_null,\n" +
                 "        '1111-01-05 17:04:42.123456+00'::timestamptz                                                           as col_timestamptz,\n" +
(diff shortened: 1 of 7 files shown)


FireboltDotNetSdk.Tests/Integration/DataTypesTest.cs
```diff
@@ -331,6 +334,199 @@ public void ExecuteStreamedQueryWithDifferentDataTypes()
...
+                    {
+                        Assert.That(standardActual, Is.EqualTo(expected), $"Mismatch at '{key}'");
+                        Assert.That(genericActual, Is.EqualTo(expected), $"Mismatch at '{key}'");
+                    });
+                }
+            }
+
+            // Also deserialize into POCO using FireboltStructNameAttribute and validate
+            var poco = reader.GetFieldValue<VariousStructPoco>(0);
+            var propByAlias = typeof(VariousStructPoco)
```
ptiurin, line 497: It would be cleaner to make this a separate test. We can always avoid setup duplication by creating a method to do this.

FireboltNETSDK/Utils/FireboltStructNameContractResolver.cs
```diff
@@ -0,0 +1,62 @@
...
+                var map = pinfo.GetCustomAttribute<FireboltStructNameAttribute>(true);
+                if (map is { Name: { Length: > 0 } name })
+                {
+                    var jp = parameters.FirstOrDefault(p => p.PropertyName == pinfo.Name);
+                    if (jp != null) jp.PropertyName = name;
+                }
+            }
+
+            return parameters;
+        }
```
ptiurin, line 44: What do we do here? Let's add a comment so it's clear

FireboltNETSDK/Client/FireboltDataReader.cs
```diff
@@ -548,5 +550,43 @@ protected override DbDataReader GetDbDataReader(int ordinal)
...
+            var value = GetValue(ordinal);
+            switch (value)
+            {
+                case T direct:
+                    return direct;
+
+                case Dictionary<string, object?> dict when typeof(T).IsAssignableFrom(typeof(Dictionary<string, object>)) ||
+                                                           typeof(IDictionary<string, object>).IsAssignableFrom(typeof(T)):
+                    return (T)(object)dict;
+
```
ptiurin, line 570: Do we need the data to be assignable both ways? I think we only care about converting from the dict into the given type `T` or am I missing something?

FireboltNETSDK/Client/FireboltDataReader.cs
```diff
@@ -548,5 +550,43 @@ protected override DbDataReader GetDbDataReader(int ordinal)
...
+                        var serializer = new JsonSerializer
+                        {
+                            ContractResolver = new FireboltStructNameContractResolver()
+                        };
+                        var token = JToken.FromObject(dict);
+                        return token.ToObject<T>(serializer)!;
+                    }
+                default:
+                    // Fallback conversion attempt for compatible primitives
+                    return (T)Convert.ChangeType(value, typeof(T));
```
ptiurin, line 582: Why don't we fallback to the parent class's `GetFieldValue`?

FireboltNETSDK/Client/FireboltDataReader.cs
bogdantruta-firebolt: Initially I had a good explanation for this but now I don't, so yes, one check should be enough :)))

FireboltDotNetSdk.Tests/Unit/FireboltDataReaderTest.cs
```diff
@@ -721,5 +721,67 @@ public void DataTableCompatibility()
...
+            Assert.That(reader2.Read(), Is.True);
+            Assert.Throws<InvalidCastException>(() => reader2.GetFieldValue<TestPoco>(0));
+        }
+
+        class TestPoco
+        {
+            [FireboltStructName("int_val")] public int IntVal { get; init; }
+            [FireboltStructName("str_val")] public string StrVal { get; init; } = string.Empty;
+            [FireboltStructName("arr_val")] public float[]? ArrVal { get; init; }
+        }
```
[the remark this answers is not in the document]
ptiurin: I meant where there are some attributes marked as `[FireboltStructName("int_val")]` and some don't. Like a mix of [these](https://github.com/firebolt-db/firebolt-net-sdk/pull/135/files#diff-af32368652d787b88f6def99ad0ad4dc20f5651c21767006b02b76e2aa38e1afR156-R177) two classes.

PR #138, opened 2025-10-01
feat: Implement GetColumnSchema

bogdantruta-firebolt:
Implemented GetSchemaTable which unblocks GetColumnSchemaAsync and GetColumnSchema.
This creates a DataTable with information about the columns


goprean (commented):
Can you check as I see that sonar reported one new issue. Until we have the quality gate failing for new issues, let's keep an eye on the sonar results to make sure we don't add tech debt.

FireboltDotNetSdk.Tests/Integration/DataTypesTest.cs
```diff
@@ -278,6 +348,60 @@ public void ExecuteReaderWithDifferentDataTypes()
...
+            var schema = await reader.GetColumnSchemaAsync();
+            Assert.That(schema, Has.Count.EqualTo(TypeList.Count));
+
+            for (var i = 0; i < schema.Count; i++)
+            {
+                Assert.Multiple(() =>
+                {
+                    Assert.That(schema[i].ColumnName, Is.EqualTo(StructAliases[i]));
+                    Assert.That(schema[i].DataType, Is.EqualTo(TypeList[i]));
+                    Assert.That(schema[i].DataTypeName, Is.EqualTo(reader.GetDataTypeName(i)));
```
goprean, line 400: I think to verify these 3 more properties would be something useful for the caller of this method to know about a column:
- if the column is null (isAllowDbNull)
- if the column has precision and scale then let's verify those as well (this would be only for double precision and decimal). You can have a new tests where you validate the scale and precision.

goprean (approved):
lgtm

